PR #1, opened 2019-06-05
MasterUserRecord API

alexeykazakov:
Initial version of MasterUserRecord API


alexeykazakov:
This PR defines go module, so, to build the repo use the following commands:
```
$ export GO111MODULE=on
$ go build ./...
```
`GO111MODULE=on` is required if run inside `$GOPATH/src` directory.

dipak-pawar:
@alexeykazakov I think we should move your comments under documentation

alexeykazakov:
@sbryzak did I miss any property?

sbryzak:
@alexeykazakov ah no, my apologies I forgot we were storing `nsLimit` etc as labels, not as part of the spec.

MatousJobanek:
> my apologies I forgot we were storing nsLimit etc as labels, not as part of the spec.

that was one of my proposals in the doc: https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit?ts=5cf5eb47#heading=h.y9t6t3s7q9vb that we could store them as parameters and not as labels when the `ProvisionedUser` is our own custom resource

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Desired state of the user record: approved|banned|deactivated
+	State string `json:"state,omitempty"`
+
+	// The list of user accounts in the member clusters which belong to this MasterUserRecord
+	UserAccounts []UserAccountEmbedded `json:"userAccounts,omitempty"`
+}
+
+type UserAccountEmbedded struct {
```
xcoulon: do we really need to suffix the type with `Embedded`?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -0,0 +1,131 @@
...
+type UserAccountEmbeddedSpec struct {
+
+	// The namespace limit name
+	NSLimit string `json:"nsLimit"`
+
+	// Embedded namespace template set
+	NSTemplateSet NSTemplateSetEmbedded `json:"nsTemplateSet"`
+}
+
+type NSTemplateSetEmbedded struct {
```
xcoulon: do we really need to suffix the type with `Embedded`?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Observed status. For example: provisioning|provisioned
+	Status string `json:"status,omitempty"`
+
+	// The status of user accounts in the member clusters which belong to this MasterUserRecord
+	UserAccounts []UserAccountStatusEmbedded `json:"userAccounts,omitempty"`
+}
+
+type UserAccountStatusEmbedded struct {
```
xcoulon: do we really need to suffix the type with `Embedded`?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
alexeykazakov: We will have UserAccount type to represent UserAccount in member. It's the same package so we need that prefix to avoid conflicts.

MatousJobanek: Yes, and it's different than the `UserAccount` itself - it also contains the `targetCluster` and `resourceVersion`

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
MatousJobanek: It's "fixed" in the new PR: https://github.com/codeready-toolchain/api/pull/3

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
MatousJobanek: Same as for the `UserAccount` - it has additional parameters

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
xcoulon: so we'll have `MasterUserRecord` CRs on the host cluster and `UserAccount` CRs on member clusters? I have to say that I'm not a big fan of this `Embedded` prefix.

Couldn't we have something like this instead, using a map to index the `UserRecord` - not `UserAcccountEmbedded` by cluster, if it's doable:

```
type MasterUserRecordSpec struct {

    // Desired state of the user record: approved|banned|deactivated
    State string `json:"state,omitempty"`

    UserAccounts map[string]UserRecord `json:"userAccounts,omitempty"`
}

type UserRecord struct {

    // The resource version of the corresponding UserAccount
    ResourceVersion string `json:"resourceVersion"

    // The namespace limit name
    NSLimit string `json:"nsLimit"`

    // Embedded namespace template set
    NSTemplateSet NSTemplateSetEmbedded `json:"nsTemplateSet"`
}
```
(also, note that here the `UserRecord` type contains the `ResourceVersion` field.)

once we have that, we can keep the `UserAccount` type for the member clusters.

WDYT?

MatousJobanek: I probably don't understand your concern. Is your concern that we mix two CRDs that will live within two separated clusters? If yes, then your solution won't help as the `NSTemplateSet` is supposed to be used only in the member as well. 
If your concern is that you just don't like the `Embedded` suffix, then we can change the `UserAccountEmbedded` to `UserRecord`, would that help?

xcoulon: @MatousJobanek yes my concern was initially about this `Embedded` suffix. I also wondered if having a map of "accounts" per cluster would help in the future, but that's another story.

xcoulon: and yes, since we are moving toward using `MasterUserRecord`, we should probably use `UserRecord` as the embedded type, for a matter of consistency

MatousJobanek: > I also wondered if having a map of "accounts" per cluster would help in the future, but that's another story.

I don't think that we will have many of the records for one user account, so I think that it won't help us much from the optimization point of view. Personally, I like more the current approach when we have the cluster identifier (either name or the address) as a value and not as a key.

> we should probably use UserRecord as the embedded type, for a matter of consistency

I think that it is a valid point, WDYT @alexeykazakov @sbryzak ? Could we use `UserRecord` instead of `UserAccountEmbedded`? From my point of view, it looks better.

alexeykazakov: UserAccountEmbedded represents a spec + additional params of a UserAccount resource which is embedded in another "parent" resource. In this case it's embedded in a MasterUserRecord. Not sure what's wrong with this name.

PR #3, opened 2019-06-11
UserAccount, NSTemplateSet and NSTemplateTier API

alexeykazakov:


pkg/apis/toolchain/v1alpha1/useraccount_types.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// UserAccountStatus defines the observed state of UserAccount
+// +k8s:openapi-gen=true
+type UserAccountStatus struct {
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Inherits the status from the corresponding NSTemplateSet
+	NSTemplateSetStatus `json:",inline"`
```
MatousJobanek: To be honest, I'm not sure about this status part that is inherited in all CRDs. I have a feeling that each of the CRD should contain the status field for the CR itself plus some info/error for the immediate successor. We should maybe change the `UserAccountStatus` (and thus also indirectly `MasterUserAccount`) to:
```go
type UserAccountStatus struct {
	Status string `json:"status,inline"`
	Error  string `json:"error,inline"`
}
```
and let only the `NsTemplateSet` keep the statuses of the particular namepaces

pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
```diff
@@ -0,0 +1,86 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// NSTemplateSetSpec defines the desired state of NSTemplateSet
+// +k8s:openapi-gen=true
+type NSTemplateSetSpec struct {
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// The name of the tier represented by this template set
+	TierName string `json:"tierName"`
```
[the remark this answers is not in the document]
alexeykazakov: Right. But I just not sure what enums exactly we will have. Let's discuss it the proposed "admin", "advanced", "basic" with PM first then add it to API.

pkg/apis/toolchain/v1alpha1/useraccount_types.go
alexeykazakov: Yeah. Maybe it's better. I've changed the status - https://github.com/codeready-toolchain/api/pull/3/commits/69fcc6f459060519a3077333045bc2e5c967533d

MatousJobanek: I meant that we should change only the `UserAccountStatus` and `MasterUserAccount` just indirectly via the "inheritance". We need to keep at least the `ResourceVersion` to know if the  `MasterUserAccount` should sync (check the updated status in `UserAccount` and `TargetCluster` for mapping it.

alexeykazakov: Right. I cleaned too much :)
https://github.com/codeready-toolchain/api/pull/6
https://github.com/codeready-toolchain/host-operator/pull/6

PR #4, opened 2019-06-12
chore: add  Makefile to build and generate

xcoulon:


xcoulon:
> LGTM. I resolved the conflict and did some minor changes.
> 
> It would be useful to add openapi generator to `make generate` too but we can do it in a separate PR.

yes, I'll add that, too

Makefile
```diff
@@ -0,0 +1,34 @@
...
+include ./make/help.mk
+include ./make/out.mk
+include ./make/go.mk
+include ./make/git.mk
+include ./make/format.mk
+include ./make/lint.mk
+
+.PHONY: build
+## Build
+build: vendor $(shell find . -path ./vendor -prune -o -name '*.go' -print)
```
dipak-pawar: Do we really to do vendor while building it. I think we don't need vendor as it's only needed by generators, which needs to be in `generate` target. or am I missing something?

make/boilerplate.go.txt
```diff
@@ -0,0 +1,16 @@
+/*
```
dipak-pawar: not sure, about why is it in `make/boilerplate.go.txt`? and also is this licence correct?

make/format.mk
```diff
@@ -0,0 +1,24 @@
+ifndef FORMAT_MK
+FORMAT_MK:=# Prevent repeated "-include".
+
+include ./make/out.mk
+
+GOFORMAT_FILES := $(shell find  . -name '*.go' | grep -vEf ./gofmt_exclude)
+
+.PHONY: check-go-format
+## Exits with an error if there are files that do not match formatting defined by gofmt
+check-go-format: ./vendor
```
dipak-pawar: not sure why do we need to do `vendor` here, as we are not going to check formatting in vendor directory

xcoulon: good point, let me remove that dependency

make/boilerplate.go.txt
xcoulon: that's the default name of the file copied from vendor/k8s.io/code-generator/hack/boilerplate.go.txt but since I was not able to force the import of this package (it has no go code), I opted for the `--go-header` flag in the `generate` goal to specify the path. The license is the same as the one in the k8s.io/code-generator pkg

Makefile
xcoulon: we don't need the `vendor` folder at all, but I believe it's helpful to have a local copy in the repository so we can easily browse the content from the IDE, which is why I've added the dependency here.

dipak-pawar: sure, but do we need do it for every build. instead of that I would have another target for dev something like `build-with-vendor`, so that whoever wants can use it

make/boilerplate.go.txt
xcoulon: I can rename this file to `go-header.txt` so the name won't feel so confusing or "hack-ish", ok?

dipak-pawar: +1 to rename

Makefile
alexeykazakov: Isn't vendor needed/used by generator(s)?
I mean yeah, for IDE we could have build-with-vendor but we would also need to use that target in make generate.

make/boilerplate.go.txt
xcoulon: done in https://github.com/codeready-toolchain/api/pull/4/commits/337edabc19122b25cd05e592386ca666c299e2a0

Makefile
xcoulon: good point @alexeykazakov, I've restored it in 2def592 (was removed in a previous commit in this PR)

xcoulon:
@alexeykazakov @dipak-pawar I decided to move on an merge this PR since @alexeykazakov already approved it, and because the `make build` goal is needed to use OpenShift CI.

PR #7, opened 2019-06-14
enable go modules in make goals

xcoulon:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,16 @@ include ./make/lint.mk
 
 .PHONY: build
 ## Build
-build: vendor $(shell find . -path ./vendor -prune -o -name '*.go' -print)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,15 +12,16 @@ include ./make/lint.mk
 
 .PHONY: build
 ## Build
-build: vendor $(shell find . -path ./vendor -prune -o -name '*.go' -print)
-	$(Q)CGO_ENABLED=0 GOARCH=amd64 GOOS=linux \
+build: $(shell find . -path ./vendor -prune -o -name '*.go' -print)
+	$(Q)CGO_ENABLED=0 GOARCH=amd64 GOOS=linux GO111MODULE=on \
```
dipak-pawar: I would export it once, instead of having exported for every command

PR #8, opened 2019-06-14
add goal to run tests in makefile

xcoulon:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,13 @@ generate: vendor
 	--bounding-dirs github.com/codeready-toolchain/api/pkg/apis "toolchain:v1alpha1" \
 	--go-header-file=make/go-header.txt
 	
-	
+.PHONY: test
+## runs the tests
+test:
+	@echo "running the tests..."
+	$(Q)GO111MODULE=on \
+	go test ./...
+
 .PHONY: clean
 ## Clean
 clean:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,7 +26,13 @@ generate: vendor
 	--bounding-dirs github.com/codeready-toolchain/api/pkg/apis "toolchain:v1alpha1" \
 	--go-header-file=make/go-header.txt
 	
-	
+.PHONY: test
+## runs the tests
+test:
+	@echo "running the tests..."
+	$(Q)GO111MODULE=on \
```
dipak-pawar: can we export `GO111MODULE=on` only once?

xcoulon: @dipak-pawar I tried to set the variable once but it didn't work so for now I'm doing this. I can revisit this part later if needed

xcoulon: but if this doesn't block, I'd like to move on so we can have the OpenShift CI in place for this repo

xcoulon: ah wait, I can do this, indeed:
```
GO111MODULE?=on
export GO111MODULE
```
let me change that

xcoulon: done in https://github.com/codeready-toolchain/api/pull/8/commits/e23614cb5689a4af4839bf9ed6aa3a16ce0285e8

PR #9, opened 2019-06-17
regenerate OpenAPI .go file and CRD files from types

xcoulon:
- generate the openapi go file
- generates the CRD files and dispatch them in the 'host-operator'
and 'member-operator' repositories, but prior to that, makes sure
that both target repositories don't have any pending changes (except
for the CRD files if they need to be generated a few times because
of refactoring)

Fixes #CRT-142


nurali-techie:
@xcoulon looks weird that from `api` project we are generating some resource and pushing to host/member operator.  Is this approached discussed / brainstormed elsewhere either with operator_team or our_team ??

xcoulon:
> @xcoulon looks weird that from `api` project we are generating some resource and pushing to host/member operator. Is this approached discussed / brainstormed elsewhere either with operator_team or our_team ??

@nurali-techie yes I've discussed about it with @MatousJobanek and @alexeykazakov. I'm open to a better suggestion if you have any.

MatousJobanek:
@nurali-techie Yes, it was discussed. 
We need to keep the api types in a separated repository, but the CRD open API shema definitions in the repository where they belong to. It is done in all OpenShift repositories/projects in the same way. 
To make the live easier, it is much better to generate the schemas from the types  and move it to the other repos using a Makefile target than doing it manually.

Makefile
```diff
@@ -10,6 +10,15 @@ include ./make/git.mk
 include ./make/format.mk
 include ./make/lint.mk
 
+# current groupname and version of the operators'API
+API_GROUPNAME=toolchain
+API_VERSION:=v1alpha1
+
+# how to dispatch the CRD files per repository (space-separated lists)
+HOST_CLUSTER_CRD_FILES:=$(API_GROUPNAME)_$(API_VERSION)_masteruserrecord.yaml $(API_GROUPNAME)_$(API_VERSION)_nstemplatetier.yaml $(API_GROUPNAME)_$(API_VERSION)_useraccount.yaml
```
MatousJobanek: UserAccount should be in Member

Makefile
```diff
@@ -18,13 +27,82 @@ build: $(shell find . -path ./vendor -prune -o -name '*.go' -print)
...
+	--output-package github.com/codeready-toolchain/api/pkg/apis/$(API_GROUPNAME)/$(API_VERSION) \
+	--output-file-base zz_generated.openapi \
+	--go-header-file=make/go-header.txt
+
+# make sure that that the `host-operator` and `member-operator` repositories exist locally 
+# and that they don't have any pending changes (except for the CRD files). 
+# The reasonning here is that when a change is made in the `api` repository, the resulting changes 
+# in the `host-operator` and `member-operator` repositories can be pushed at the same time on GitHub, 
+# without having to wait for some other feature or fix to be completed.
+# TODO: we could even go further and checkout new branches from `master` in the 'host-operator'
```
MatousJobanek: that sounds good :+1:

Makefile
```diff
@@ -18,13 +27,82 @@ build: $(shell find . -path ./vendor -prune -o -name '*.go' -print)
...
+# and 'member-operator` repositories and give them the name of the current branch in this repo.
+# The developer would have 3 branches with the same name and could then push to GitHub at the 
+# same time...
+host_repo_status := $(shell cd ../host-operator && git status -s | grep -v deploy/crds)
+member_repo_status := $(shell cd ../member-operator && git status -s | grep -v deploy/crds)
+
+PHONY: host-operator
```
MatousJobanek: could we change the name for example to `prepare-host-operator-repo` so it is clear what the target does? Analogically for the `member-operator`

Makefile
```diff
@@ -18,13 +27,82 @@ build: $(shell find . -path ./vendor -prune -o -name '*.go' -print)
...
+host-operator: ../host-operator
+ifdef host_repo_status
+	@echo "The local '../host-operator' repository has pending changes. Please stash them or commit them, first."
+	@exit 1
+endif
+ifneq ($(wildcard ../host-operator/deploy/crds/*.yaml),)
+	@-rm ../host-operator/deploy/crds/*.yaml
```
MatousJobanek: The directory cannot be cleaned (you cannot remove all CRDs) because it will contain for example also KubeFedCluster CRD that is not generated from our API repo: https://github.com/codeready-toolchain/member-operator/pull/15/files#diff-38b5ba4a46c45971ace7a2b8fae8fa77

Makefile
```diff
@@ -18,13 +27,82 @@ build: $(shell find . -path ./vendor -prune -o -name '*.go' -print)
...
+	@-rm ../member-operator/deploy/crds/*.yaml
+else
+	@-mkdir -p ../member-operator/deploy/crds
+endif
+
+
+.PHONY: generate-crds
+generate-crds: host-operator member-operator
+	@echo "Re-generating the CRD files..."
+	$(Q)go run $(shell pwd)/vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd \
```
MatousJobanek: This means that it requires the vendor directory, but this repository uses the go module. Could you please add another target that would generate the vendor directory and would be executed as the first one?

xcoulon:
> Looks awesome.
> Apart from the in-line comments, I have one request - could you please add some description to the README so people with the same [gold-fish memory](http://2.bp.blogspot.com/-yHK_nOwLiKA/VcdylNAaR7I/AAAAAAAAAoY/2ZFeNEAn4vg/s1600/cartoon1.gif) as I have don't forget which target should be used and how, thanks :-)

damn'it, I already did that but I did not realize that the `git push` command failed on my machine earlier today (problem with the Wi-Fi). Let me push it now

Makefile
xcoulon: ok, so I suggest that I delete all files except the `*kubefed*` ones. The reason for doing this is that in case of refactoring, the `.yaml` file may change and we don't want to keep the old files. Does this sound ok for you, or should I just skip this step and let the developers manually remove the obsolete files?

Makefile
xcoulon: done in 987e79c

Makefile
xcoulon: done in 987e79c. Feel free to object ;)

Makefile
xcoulon: done in 987e79c

Makefile
xcoulon: fixed in 987e79c

xcoulon:
@MatousJobanek I believe I addressed all your comments. Also, I removed the "trick" to avoid multiple "includes" of the same sub-makefiles by just doing an `include ./make/*.mk` in the Makefile. I believe it's simpler/cleaner and it won't require any extra maintenance if we add more sub-makefiles (Also, I got some errors with the `ifndef`/`endif`  trick)

Could you give a second look at this PR , please?

Makefile
MatousJobanek: I see, but the developer will need to do manual step anyway - to change the target that deploys the CRDs. But I agree that this is safer. 
We can keep it as it is - excluding one CRD is fine and I doubt that there will be more of them :-)

xcoulon:
thanks for the review, all! 🙌

stevekuznetsov:
@xcoulon @alexeykazakov you've enabled merge automation but in a way that it can't actually do anything:

```
{
  "base-sha": "c1c49433bba5131f966ebbba5d3992d34d97cfb0",
  "branch": "master",
  "component": "tide",
  "controller": "sync",
  "error": "Tide needs to be configured to use the 'rebase' merge method for this repo or the repo needs to allow merge commits: Merge commits are not allowed on this repository.",
  "file": "prow/tide/tide.go:896",
  "func": "k8s.io/test-infra/prow/tide.(*Controller).mergePRs",
  "level": "error",
  "merge-targets": [
    9
  ],
  "msg": "Merge failed.",
  "org": "codeready-toolchain",
  "pr": 9,
  "repo": "api",
  "sha": "987e79c72cf30009f53ff086a50f3ea0a8e97cd8",
  "time": "2019-06-18T15:29:16Z"
}
```

If you want the merge automation to do the merge for you please change your strategy to `rebase` [here](https://github.com/openshift/release/blob/039a8f1a9516aa6baeab544d8c711b570c697667/cluster/ci/config/prow/config.yaml#L80).

xcoulon:
@stevekuznetsov thanks for reporting this error, but I'm not sure I fully understand what how the merge automation works. Would a PR be automatically merged if it get one `/lgtm` comment? I'm not sure this is what we want. Also, we (usually) use the `squash and merge` option on GitHub. Is this also doable with Prow?

stevekuznetsov:
The configuration that you added [here](https://github.com/openshift/release/blob/946119a3a810feaf7801bd0667a00cb10bc64d8d/cluster/ci/config/prow/config.yaml#L500):

```yaml
  - repos:
    - codeready-toolchain/host-operator
    - codeready-toolchain/api
    - codeready-toolchain/member-operator
    labels:
    - lgtm
    - approved
    missingLabels:
    - needs-rebase
    - do-not-merge/blocked-paths
    - do-not-merge/hold
    - do-not-merge/work-in-progress
    - do-not-merge/invalid-owners-file
```

Will wait for `lgtm` and `approved` labels to exist before considering a job for merge. Once that is fulfilled, tests must pass against the most recent version of the PR and the branch it's targeting before a merge will happen.

> Also, we (usually) use the squash and merge option on GitHub. Is this also doable with Prow?

Yes, as noted in the first comment there is a field to configure this.

xcoulon:
@stevekuznetsov ok, I've enabled the "merge commits" as suggested in the error message of Prow, and I've also opened https://github.com/openshift/release/pull/4090 to set the merge method to `squash` on all our 3 repositories.

PR #11, opened 2019-06-26
add useracc status const

nurali-techie:
Two status constant added for UserAccount.  "provisioning" and "provisioned".
This will be used by member-operator/UserAccCtrl in PR https://github.com/codeready-toolchain/member-operator/pull/20

### DIFF
--- a/pkg/apis/toolchain/v1alpha1/useraccount_types.go
+++ b/pkg/apis/toolchain/v1alpha1/useraccount_types.go
@@ -6,6 +6,11 @@ import (
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+const (
+	StatusProvisioning string = "provisioning"
+	StatusProvisioned  string = "provisioned"
+)
+
 // UserAccountSpec defines the desired state of UserAccount
 // +k8s:openapi-gen=true
 type UserAccountSpec struct {


pkg/apis/toolchain/v1alpha1/useraccount_types.go
```diff
@@ -6,6 +6,11 @@ import (
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+const (
+	StatusProvisioning string = "provisioning"
+	StatusProvisioned  string = "provisioned"
```
MatousJobanek: May I know why defining it as a specific type of string? eg:
```go 
type StatusUserAccount string
const (
	StatusProvisioning StatusUserAccount = "provisioning"
	StatusProvisioned  StatusUserAccount = "provisioned"
)
```
With this it is much easier to check which values (statuses) are defined for the object.

alexeykazakov: Yeah. Good point @MatousJobanek

nurali-techie: @MatousJobanek I have thought of same thing but problem is that the status type is `string` [link](https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/useraccount_types.go#L38).  Once we have type like `StatusUserAccount` and assignment to string or comparison with string need to cast status.  Check this https://play.golang.org/p/PpH4ZrbcInh

alexeykazakov: We can change the status type in UserAccount API from `string` to `StatusUserAccount` - https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/useraccount_types.go#L38

nurali-techie: Yup, that's the thing I want to try to make use of `StatusUserAccount` type.

MatousJobanek: in other words something like this, right? https://github.com/codeready-toolchain/api/pull/13

PR #13, opened 2019-06-27
feat: introduce StatusUserAccount type

MatousJobanek:
to make the comparison/navigation easier

### DIFF
--- a/pkg/apis/toolchain/v1alpha1/useraccount_types.go
+++ b/pkg/apis/toolchain/v1alpha1/useraccount_types.go
@@ -6,9 +6,10 @@ import (
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type StatusUserAccount string
 const (
-	StatusProvisioning string = "provisioning"
-	StatusProvisioned  string = "provisioned"
+	StatusProvisioning StatusUserAccount = "provisioning"
+	StatusProvisioned  StatusUserAccount = "provisioned"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/apis/toolchain/v1alpha1/useraccount_types.go
```diff
@@ -35,7 +36,7 @@ type UserAccountStatus struct {
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 
 	// Observed status. For example: provisioning|provisioned
-	Status string `json:"status,omitempty"`
+	Status StatusUserAccount `json:"status,omitempty"`
```
nurali-techie: hope `make generate` target run (to generate stuff in member/host operator) and things are validated to make sure member/host operator are fine with the changes ??

xcoulon: @nurali-techie `make generate` dispatches the CRD files in the `member-operator` and `host-operator` repositories, but it won't run the `make build` or `make test` in those repo.

nurali-techie: I am talking about manual_validation after running `make generate`.

MatousJobanek: I probably don't understand your question. `make generate` just generates and dispatches the CRDs - that doesn't have any impact on the code... Anyway, this change that is introduced in this PR shouldn't change the CRD anyhow - the value is still string (verified locally ;-)).

PR #15, opened 2019-07-03
Add conditions to UserAccount & MasterUserRecord statuses

alexeykazakov:
Also adding new deprovisioned and disabled fields to spec.


pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -4,6 +4,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// These are valid conditions of a MasterUserRecord
+const (
+	// MasterUserRecordProvisioning means the Master User Record is being provisioned
+	MasterUserRecordProvisioning ConditionType = "Provisioning"
+	// MasterUserRecordUserAccountNotReady means the User Account failed to be provisioned
+	MasterUserRecordUserAccountNotReady ConditionType = "UserAccountNotReady"
+	// MasterUserRecordReady means the Master User Record failed to be provisioned
```
MatousJobanek: This comment is either wrong or I don't understand the purpose of the `Ready` condition - if for `Ready` type the status is set to `True` then it should mean that the `UserAccount` is ready, shouldn't it?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -4,6 +4,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// These are valid conditions of a MasterUserRecord
+const (
+	// MasterUserRecordProvisioning means the Master User Record is being provisioned
+	MasterUserRecordProvisioning ConditionType = "Provisioning"
+	// MasterUserRecordUserAccountNotReady means the User Account failed to be provisioned
+	MasterUserRecordUserAccountNotReady ConditionType = "UserAccountNotReady"
```
MatousJobanek: Why do we have `UserAccountNotReady` and not `UserAccountReady` - analogically to the `Ready` condition?
BTW, we do we have specific `UserAccountNotReady` for the `UserAccount` in MUR? this information is kept in the status `UserAccountStatusEmbedded` anyway. In addition, the `UserAccountNotReady` doesn't say which `UserAccount` (in which cluster) this condition refers to.

pkg/apis/toolchain/v1alpha1/useraccount_types.go
```diff
@@ -6,11 +6,20 @@ import (
...
+	UserAccountProvisioning ConditionType = "Provisioning"
+	// UserAccountUserNotReady means the User failed to be created
+	UserAccountUserNotReady ConditionType = "UserNotReady"
+	// UserAccountIdentityNotReady means the Identity failed to be created
+	UserAccountIdentityNotReady ConditionType = "IdentityNotReady"
+	// UserAccountUserIdentityMappingNotReady means the User Identity Mapping failed to be created
+	UserAccountUserIdentityMappingNotReady ConditionType = "UserIdentityMappingNotReady"
+	// UserAccountNSTemplateSetNotReady means the NSTemplateSet failed to be provisioned
+	UserAccountNSTemplateSetNotReady ConditionType = "NSTemplateSetNotReady"
+	// UserAccountReady means the User Account failed to be provisioned
```
MatousJobanek: I have the same concern as for the MUR conditions. But now I understand why the conditions for secondary resources are `NotReady` - so they don't have to be set when the creation is successful. However, I'm not sure that we need separate conditions for User, Identity and IdentityMapping resources? In case that creation/update of one of the resources fails, then the provisioning doesn't go further and keeps reconciling and trying to create still the same resource. In other words, there should be only one problem that prevents controller from proper User provisioning at one time. This error can be set as reason/message field of the UserAccountReady condition while the status of the condition would be set to `False`

MatousJobanek: I guess that the only condition that would be MUR controller interested in is the `UserAccountReady`, correct?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
alexeykazakov: This is wrong :) Copy-paste issue.

alexeykazakov: https://github.com/codeready-toolchain/api/pull/17

PR #20, opened 2019-07-23
Relation between UserAccount and NSTeplSet

nurali-techie:
For each UserAccount there will be one NSTemplateSet object.  To established this relationship we have added `NSTemplateSetName` attribute with type `corev1.LocalObjectReference`.  `NSTemplateSetName` will be set with name of NSTemplateSet object created for this UserAccount.  As both UserAccount and NSTemplateSet belong to same namespace, we no need to use `corev1.ObjectReference` which also provide attribute to capture Namespace.

Here, relationship is only about referring name, the idea is to detach both object so UserAccount reconciler has only responsibility to create NSTemplateSet.  There will be separate controller for NSTemplateSet entity which takes care of reconciling NSTemplateSet and related entity (for ex, namespace).


pkg/apis/toolchain/v1alpha1/useraccount_types.go
```diff
@@ -24,8 +25,12 @@ type UserAccountSpec struct {
 	// The namespace limit name
 	NSLimit string `json:"nsLimit"`
 
-	// Namespace template set
-	NSTemplateSet NSTemplateSetSpec `json:"nsTemplateSet"`
+	// The name of the tier represented by this template set
+	TierName string `json:"tierName"`
```
xcoulon: the detailed design doc mentions `NSTemplateTier`. Should you rename this field to align with the doc?

nurali-techie:
changes not needed.

PR #25, opened 2019-08-12
fix: register KubeFed types

MatousJobanek:
Register KubeFed types with the scheme

### DIFF
--- a/pkg/apis/addtoscheme_kubefed_v1beta1.go
+++ b/pkg/apis/addtoscheme_kubefed_v1beta1.go
@@ -1,0 +1,8 @@
+package apis
+
+import "sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
+
+func init() {
+	// Register the KubeFed types with the Scheme so the components can map objects to GroupVersionKinds and back
+	AddToSchemes = append(AddToSchemes, v1beta1.SchemeBuilder.AddToScheme)
+}
--- a/go.mod
+++ b/go.mod
@@ -51,6 +51,7 @@ require (
 	k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208
 	sigs.k8s.io/controller-runtime v0.1.10
 	sigs.k8s.io/controller-tools v0.1.12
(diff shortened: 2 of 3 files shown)


pkg/apis/addtoscheme_kubefed_v1beta1.go
```diff
@@ -0,0 +1,8 @@
+package apis
+
+import "sigs.k8s.io/kubefed/pkg/apis/core/v1beta1"
+
+func init() {
+	// Register the KubeFed types with the Scheme so the components can map objects to GroupVersionKinds and back
+	AddToSchemes = append(AddToSchemes, v1beta1.SchemeBuilder.AddToScheme)
+}
```
nurali-techie: This changes looks fine but I wanted to bring the change where rather adding one_new_file for each_new_scheme used .. we can just add new_scheme to `apis.go` file [link](https://github.com/codeready-toolchain/api/blob/master/pkg/apis/apis.go) directly.  Like we can just move this `init()` function to `apis.go` file.  Even we can delete `addtoscheme_toolchain_v1alpha1.go` file [link](https://github.com/codeready-toolchain/api/blob/master/pkg/apis/addtoscheme_toolchain_v1alpha1.go) and add its scheme to this `apis.go` file.

In short, one place (`apis.go`) to register all scheme rather one file for each scheme :-)

MatousJobanek: I was thinking about it too. Actually, that was my original code (I mean for the case of KubeFedCluster), but then I decided to put it into a separated file. The reason was to keep the consistency with the `addtoscheme_toolchain_v1alpha1.go` - add the scheme in a separated file (one scheme per file). 
If you want to put everything into one file, then feel free to do it. The purpose of this PR was to register KubeFed scheme

MatousJobanek: btw, the KubeFedCluster is deployed as part of our operators (in the same way as the other CRDs are), that's why I used the same approach as is used for toolchain types.

PR #26, opened 2019-08-12
fix broken build as python3.6 is retired

dipak-pawar:
We need `python3.6`, for `yamllint` which needs to formal yaml files in lint goal. We have issues while installing `python3.6` as it's retired from centos7. you can find  issue link https://bugzilla.redhat.com/show_bug.cgi?id=1739804.

Now the workaround which we are using here is  `IUS`, which stands for Inline with Upstream Stable rpm provider. I have tested this and it worked. it's just another way to install python3 rpm from another repo.

Found from how to install python3 blog https://www.digitalocean.com/community/tutorials/how-to-install-python-3-and-set-up-a-local-programming-environment-on-centos-7

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -13,6 +13,11 @@ ENV LANG=en_US.utf8 \
 
 ARG GO_PACKAGE_PATH=github.com/codeready-toolchain/api
 
+# this is workaround till we fix https://bugzilla.redhat.com/show_bug.cgi?id=1739804
+RUN yum -y install https://centos7.iuscommunity.org/ius-release.rpm \
+    python36u \
+    python36-devel
+
 RUN yum install epel-release -y \
     && yum install --enablerepo=centosplus install -y --quiet \
     findutils \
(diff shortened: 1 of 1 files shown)


openshift-ci/Dockerfile.tools
```diff
@@ -13,6 +13,11 @@ ENV LANG=en_US.utf8 \
 
 ARG GO_PACKAGE_PATH=github.com/codeready-toolchain/api
 
+# this is workaround till we fix https://bugzilla.redhat.com/show_bug.cgi?id=1739804
+RUN yum -y install https://centos7.iuscommunity.org/ius-release.rpm \
+    python36u \
+    python36-devel
```
nurali-techie: Is this workaround mentioned somewhere ?? (just to make sure it's tried and tested)

Also I see comment in bugzilla issue that they are unretiring the packages (which were retired previously)[comment_link](https://bugzilla.redhat.com/show_bug.cgi?id=1739804#c15), if so and if we are getting this issue fixed then can't we wait for sometime (day or two) ??

It would be great if you can fill desc with your understanding of what was the issue, how this will fix it.

dipak-pawar: @nurali-techie https://github.com/codeready-toolchain/api/pull/26#issue-306435201. I don't see any specific timeline like, we can unretire it in 2 days. Also I don't feel we should wait for 2 days to start fixing this kind of smaller issue if we don't have solution from unstream.

Also, we can easily revert this change, once we have this fix in place

nurali-techie: @dipak-pawar thanks for the details.  Yes, fix looks fine and simple.  Not sure if we need to apply this fix for other projects too (host, member) ??

dipak-pawar:
closing this as these packages are back

PR #27, opened 2019-08-13
Modify condition types for UserSignup

sbryzak:
This PR removes both the "PendingApproval" and "Provisioning" condition types from the UserSignup type, and adds the "Approved" condition type.


nurali-techie:
@sbryzak will there be any change for this attribute `UserSignupSpec.Approved` [code_link](https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/usersignup_types.go#L36) like this will stay or removed ??

alexeykazakov:
@nurali-techie we still need to keep `Approved` field. That field in spec is to be used by clients to indicate that the UserSignup has been manually approved by an admin. And this PR introduces a new Condition Type name for status. So it's desired state vs. actual status.

PR #30, opened 2019-08-30
fix printcolumns tags

xcoulon:
in the generated yaml files, the 'JSONPath' field would
contain backslashes before the quotes, which would cause
errors and prevent the field from being displayed in the
`oc get ...` commands


xcoulon:
thanks @alexeykazakov

PR #32, opened 2019-09-03
support ready condition for nstemplateset 

nurali-techie:
## Description
support ready condition for nstemplateset.

## Checks

    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/53


pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
```diff
@@ -34,36 +34,22 @@ type Namespace struct {
...
-	Status string `json:"status,omitempty"`
-
-	// The error message in case of failed status
-	Error string `json:"error,omitempty"`
+	// Conditions is an array of current NSTemplateSet conditions
+	// Supported condition types: ConditionReady
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
xcoulon: so what happens to the `NamespaceStatus`? is it just removed (ie, it's obsolete) or is it going to be replaced by something different?

nurali-techie: `NamespaceStatus` replaced with `Condition`.  Currently, we will be having only `Ready` condition, just like UserAccount.  Check design doc [link](https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit#heading=h.autst8l8rdnd) for more details on `Ready` condition.

MatousJobanek: For now, only one condition is fine. However, it could limit us in the future - in case we wanted to run provisioning/update in multiple controllers in parallel - in that case, we will need to introduce one condition per namespace. But let's start with simple things first

PR #34, opened 2019-09-09
added CompareTo method to NSTmplSetSpec comparision

nurali-techie:
## Description
added CompareTo method to NSTmplSetSpec comparision.

## Checks

2. Does `make generate` change anything in other projects (host-operator, member-operator)? **[no]**

    - host-operator: 
    - member-operator:


pkg/apis/toolchain/v1alpha1/types_extension.go
```diff
@@ -0,0 +1,34 @@
+package v1alpha1
```
nurali-techie: One file `types_extension.go` where extra methods related to all different types can go.

pkg/apis/toolchain/v1alpha1/types_extension.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"reflect"
+)
+
+func (first *NSTemplateSetSpec) CompareTo(second *NSTemplateSetSpec) bool {
```
xcoulon: could you add a comment on this public method?

pkg/apis/toolchain/v1alpha1/types_extension_test.go
```diff
@@ -0,0 +1,109 @@
...
+func TestNSTemplateSetSpecCompareTo(t *testing.T) {
+	tables := []struct {
+		name   string
+		first  *NSTemplateSetSpec
+		second *NSTemplateSetSpec
+		want   bool
+	}{
+		{
+			name: "both_same",
+			first: &NSTemplateSetSpec{
```
xcoulon: could add some test cases with some template not being an empty string, too?

pkg/apis/toolchain/v1alpha1/types_extension.go
nurali-techie: done https://github.com/codeready-toolchain/api/pull/34/commits/9602d2a3af224f7899f259c52cdfae0f6de21c76

pkg/apis/toolchain/v1alpha1/types_extension_test.go
nurali-techie: done https://github.com/codeready-toolchain/api/pull/34/commits/364c155cd17844282c886e541776c68541adf062

pkg/apis/toolchain/v1alpha1/types_extension.go
xcoulon: actually, why not having this `CompareTo` method in https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/nstemplateset_types.go ?

nurali-techie: I guess `xxx_types.go` stay small and easy to understand types design.  These `CompareTo()` and other custom methods (added in future) will not of much concern and stay out of actual types definition to have has small/short type definition which is easy to grasp.

In any case, if you want, I can move `CompareTo` to `nstemplateset_types.go`.

xcoulon: well, I believe it makes sense ti have this method in `nstemplateset_types.go`, next to the type definition, since the function has the `NSTemplateSetSpec` type as a receiver.

xcoulon: it would make sense to have pure common/utility functions in another file, though

nurali-techie: done https://github.com/codeready-toolchain/api/pull/34/commits/352b9f4814c418a5f8cc243a84ad3aaa3a8c08f0

pkg/apis/toolchain/v1alpha1/nstemplateset_types_test.go
```diff
@@ -0,0 +1,147 @@
...
+				Namespaces: []Namespace{
+					{Type: "dev", Revision: "rev1", Template: tmpl1dev},
+					{Type: "code", Revision: "rev1", Template: tmpl1code},
+				},
+			},
+			second: &NSTemplateSetSpec{
+				TierName: "basic",
+				Namespaces: []Namespace{
+					{Type: "code", Revision: "rev1", Template: tmpl2code},
+					{Type: "dev", Revision: "rev1", Template: tmpl2dev},
```
xcoulon: sp can you test that the 2 NSTemplateSetSpec are different if one of their `namespace.Template` is different?

nurali-techie: done https://github.com/codeready-toolchain/api/pull/34/commits/9023f0dc41d66d34dcaff1d04589efc8d6885b9e

nurali-techie:
@xcoulon thanks for your review.

nurali-techie:
lgtm

pkg/apis/toolchain/v1alpha1/nstemplateset_types_test.go
```diff
@@ -0,0 +1,166 @@
...
+	---
+	apiVersion: v1
+	kind: Template
+	metadata:
+		labels:
+		project: codeready-toolchain
+		name: codeready-toolchain-code`
+)
+
+func TestNSTemplateSetSpecCompareTo(t *testing.T) {
```
MatousJobanek: Could you please also add the case with totally identical templates (having same order) and check that they equal using `reflect.DeepEqual`. Just to ensure that when a new param is added to NsTemplateSet, then the test will fail...?

pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
```diff
@@ -70,3 +72,33 @@ type NSTemplateSetList struct {
 func init() {
 	SchemeBuilder.Register(&NSTemplateSet{}, &NSTemplateSetList{})
 }
+
+// CompareTo compares given second spec with this spec
+func (first *NSTemplateSetSpec) CompareTo(second *NSTemplateSetSpec) bool {
```
MatousJobanek: It's a bit weird that it takes a pointer when it "only" reads the values....

pkg/apis/toolchain/v1alpha1/nstemplateset_types_test.go
nurali-techie: @MatousJobanek I am not getting this scenario, can you please detailed out with example.

pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
nurali-techie: It just a copy/paste mistake, I will change it.

pkg/apis/toolchain/v1alpha1/nstemplateset_types_test.go
MatousJobanek: never mind - I didn't realize that it wouldn't solve the problem either. So no need to add any additional test :-) 
Sorry for confusions ;-)

pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
nurali-techie: I revisited, the idea is NOT to copy the whole struct locally for `CompareTo()` so used pointer for pass_by_ref.

If we consider java, it's always pointer (pass_by_ref) even for `Object.equal()` method.  In Go, we have both options and to avoid copy whole structure, I preferred pointer.

pkg/apis/toolchain/v1alpha1/nstemplateset_types_test.go
nurali-techie: Kool, got the confusion ;-)

pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
xcoulon: Go is not like Java, you can (and should) pass by value when the parameter is not going to be changed by the called function, which is the case here.

nurali-techie: @xcoulon not sure that's always the case.  I come across this https://github.com/golang/go/wiki/CodeReviewComments#receiver-type which is particularly for `ReceiverType`.  Check point 4:

> If the receiver is a large struct or array, a pointer receiver is more efficient. How large is large? Assume it's equivalent to passing all its elements as arguments to the method. If that feels too large, it's also too large for the receiver.

But this is for `ReceiverType`, so I can't arg same for function input param using this point.

Anyway, I will change it to pass_by_value :+1:

nurali-techie: Fixed https://github.com/codeready-toolchain/api/pull/35

PR #36, opened 2019-09-17
Pin to kube 1.14 & cleanup

alexeykazakov:
- Pin to kube 1.14 (OpenShift 4.2 will be based on Kube 1.14)
- `go mod tidy`
- Still using `sigs.k8s.io/controller-tools v0.1.12` instead of `v.0.2.0` because our code generation doesn't support the newer version. We will need to fix it before upgrading to 0.2.0


go.mod
```diff
@@ -3,69 +3,46 @@ module github.com/codeready-toolchain/api
...
 	k8s.io/apiextensions-apiserver v0.0.0-20190610170453-8677a5df3599 // indirect
 	k8s.io/apimachinery v0.0.0-20190607205628-5fbcd19f360b
-	k8s.io/code-generator v0.0.0-20190531131525-17d711082421
+	k8s.io/code-generator v0.0.0-20181117043124-c2090bec4d9b
 	k8s.io/gengo v0.0.0-20190327210449-e17681d19d3a
 	k8s.io/klog v0.3.3 // indirect
 	k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208
-	sigs.k8s.io/controller-runtime v0.1.10
+	sigs.k8s.io/controller-runtime v0.2.0
+	// TODO support to the newer controller-tools in pkg/types and makefile before upgrading it
```
MatousJobanek: Could you please create a Jira issue and ideally link it here so we won't forget? 
Thanks

alexeykazakov: Good point. Here is the jira - https://jira.coreos.com/browse/CRT-254

PR #47, opened 2019-10-16
configure golangci-lint new timeout

xcoulon:
## Description
Configure the `timeout` value to 10min in `.golangci.yml`

## Checks
1. Have you run `make generate` target? **no**

2. Does `make generate` change anything in other projects (host-operator, member-operator)? **no** 

## Notes

fixes CRT-298

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,10 @@
+service:
+  project-path: github.com/codeready-toolchain/api
+
+run:
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  timeout: 10m
+
 linters:
   enable:
     - gofmt


.golangci.yml
```diff
@@ -1,3 +1,10 @@
+service:
+  project-path: github.com/codeready-toolchain/api
+
+run:
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  timeout: 10m
```
dipak-pawar: I assume we don't use this timeout in `lint-go-code` here. Can you please confirm @xcoulon?

xcoulon: @dipak-pawar this is the new recommended way to configure the timeout (as shown in https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml#L10) since the `deadline` CLI arg is now deprecated. This `run.timeout` value should be used by the `golangci-lint` command as part of the `make lint`. 

Does this answer your question?

dipak-pawar: @xcoulon sure, what I am asking here is we don't configure this timeout in make target `lint-go-code`  -> 
`$(Q)${GOPATH}/bin/golangci-lint ${V_FLAG} run --deadline=10` for this repository like other repository. I have seen, you have remove `--deadline` flag from other repos. just wondering do we need to do the same for here?

dipak-pawar: no worries, we are not using `--deadline` flag for this repository, we are good.

xcoulon: ah ok, sorry for the confusion. And yes, in this repo ww did not use the `--deadline` flag, hence the lack of changes in `make/link.mk`

PR #49, opened 2019-10-29
CRT-330 remove CompliantUsername field from UserSignup.Spec

sbryzak:
Signed-off-by: Shane Bryzak <sbryzak@gmail.com>

## Description
This PR removes the CompliantUsername field from UserSignup spec.  The registration service will be refactored, and the username transformation will now take place in [UserSignupController](https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/usersignup/usersignup_controller.go).  Because of this change, the DNS-1123-compliant username will now be stored in a resource label instead, as the controller may not modify the resource spec.

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/93
    - member-operator: no significant changes


alexeykazakov:
Looks good overall but we will need to create a PR with updated e2e: https://github.com/codeready-toolchain/toolchain-e2e/blob/672242d6ea287b38a48404682beddecbdb98acc3/test/e2e/usersignup_test.go#L63

MatousJobanek:
Don't forget that we also need to know what the final MUR name for the particular UserSignup is - we should add that to the status part - it's described in this issue: https://jira.coreos.com/browse/CRT-328

sbryzak:
@alexeykazakov updated e2e tests draft PR is here: https://github.com/codeready-toolchain/toolchain-e2e/pull/29

sbryzak:
@MatousJobanek to store the MUR name, we decided that the UserSignup controller will now set a label value in UserSignup to reflect the compliant username (which will be the same as the MUR name).

MatousJobanek:
> we decided that the UserSignup controller will now set a label value in UserSignup to reflect the compliant username 

May I know the reasoning for using a label? For me it looks more logical to put it into the status as it's the result of the overall processing...

sbryzak:
> May I know the reasoning for using a label? For me it looks more logical to put it into the status as it's the result of the overall processing...

I know that you can query resources by label value, can you do the same with the status?  I couldn't find anything in the Kubernetes docs or any examples online that suggest this is supported.

alexeykazakov:
Only labels and name can be used as a selector in Custom Resources atm in K8s. So, having complaint name as a label might be useful.
But yeah... adding a label for the primary resource from the controller might be not the best thing to do... Status would be a better place.
Not sure what's more important here. Being able to query UserSignups by the complaint name (might be useful for debugging at least) or following the best practices...

MatousJobanek:
But why we would need to have such a label (as a selector) in the first place? I don't really think that we will use that selector because:

* we don't know what the resulting complaint name is without looking at the MUR name
* when we have the MUR or UserAccount, then we can easily take the `userID` which is same as the name of the `UserSignup`

So, if we want to navigate from `UserSignup` to MUR then we can take the resource and check the status
If we want to navigate from MUR to `UserSignup` then we can take the `userID`

alexeykazakov:
Let’s say we have some resource somewhere deep in kube which behaves badly
and we want to investigate. Let’s say it’s a NSTemplateSet. So we can just
look up all resources using the transformed name as a selector across all
the clusters and it would give as all relevant resources including Signups.
No need to handle them separately.
But I’m not sure if it’s worth it.

sbryzak:
One other thing I would like to point out is that Kubernetes already establishes the notion of reserved labels, see https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/.  Based on this I would modify our own reserved label keys to have a set prefix.

pkg/apis/toolchain/v1alpha1/usersignup_types.go
```diff
@@ -54,6 +48,10 @@ type UserSignupStatus struct {
 	// +patchStrategy=merge
 	// +listType
 	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// CompliantUsername is used to store the transformed, DNS-1123 compliant username
+	// +optional
+	CompliantUsername string `json:"compliantUSername,omitempty"`
```
alexeykazakov:
```suggestion
	CompliantUsername string `json:"compliantUsername,omitempty"`
```

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -92,6 +92,8 @@ type UserAccountStatusEmbedded struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=mur
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.conditions[?(@.type=="Ready")].status"
```
alexeykazakov:
```suggestion
// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].status`
```

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -92,6 +92,8 @@ type UserAccountStatusEmbedded struct {
...
+// +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=".status.conditions[?(@.type=="Ready")].reason"
```
alexeykazakov:
```suggestion
// +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
```

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -12,6 +12,8 @@ const (
 	MasterUserRecordUserAccountNotReady ConditionType = "UserAccountNotReady"
 	// MasterUserRecordReady means the Master User Record provisioning succeeded
 	MasterUserRecordReady ConditionType = "Ready"
+
+	MasterUserRecordUserIDLabelKey = "v1.toolchain/user-id"
```
alexeykazakov: If you want to version the labels then shouldn't it be `v1alpha1`? Just to align it with the API version.
Also maybe worth using the full group `toolchain.dev.openshift.com` instead of short `toolchain`?
WDYT? But I'm fine with `toolchain` too.

PR #59, opened 2019-11-14
Add details about cluster for UserAccounts in MUR

alexeykazakov:
Required for https://jira.coreos.com/browse/CRT-348

## Description
This PR adds more details about target clusters for UserAccounts in MasterUserRecord status:
- API Endpoint
- Console URL

It's required because we need to show the console URL to users in UI.

## Checks

yes

yes

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/104


pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -86,13 +91,25 @@ type UserAccountStatusEmbedded struct {
...
+}
+
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // MasterUserRecord is the Schema for the masteruserrecords API
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:scope=Namespaced
 // +kubebuilder:resource:shortName=mur
+// +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=`.spec.userAccounts[].targetCluster`
```
MatousJobanek: shouldn't that user rather the `Cluster` param/object instead of the `targetCluster` string when it's deprecated?
```suggestion
// +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=`.spec.userAccounts[].cluster.name`
```

alexeykazakov: This PR doesn't change the Target Cluster (string) from UserAccount *Spec*: `.spec.userAccounts[].targetCluster`. It basically says that User Account should be provisioned on that particular cluster name.

What is changed here is the status of UserAccount. When it's provisioned it now has a detailed cluster object with URLs.

Does it make sense to you?

alexeykazakov:
> Great improvement! I see that the generated code in the host-operator looks fine, but could you please modify also the example and regenerate the CSV one more time with the changed example?

Same explanation as in https://github.com/codeready-toolchain/api/pull/59#discussion_r346376343
In example we have only spec (no status) which is not change by this PR.

MatousJobanek:
Yeah, I missed that it's the status part. :-) so no changes in example is needed

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
MatousJobanek: Yes, you are right. Sorry for confusion

PR #61, opened 2019-11-28
feat(CRT-352): create RegistrationService CRD

MatousJobanek:
## Description
It will be used as a holder of reg-service configuration as well as a triggerrer of the actual deployment.
The params that are in the CRD have been taken from the ConfigMap used by the registration service

## Checks

**yes**

**yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/108


pkg/apis/toolchain/v1alpha1/registrationservice_types.go
```diff
@@ -0,0 +1,83 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
alexeykazakov: We can delete this comment.

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+// RegistrationServiceSpec defines the desired state of RegistrationService
+// +k8s:openapi-gen=true
+type RegistrationServiceSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// The image identifies which image of the registration service should be used for a deployment
+	Image string `json:"version"`
```
alexeykazakov:
```suggestion
	Image string `json:"image"`
```

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// The number of replicas of the deployed registration service
+	Replicas int `json:"replicas,omitempty"`
```
alexeykazakov:
```suggestion
	// +optional
	Replicas int `json:"replicas,omitempty"`
```

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// The environment identifies which mode the registration service should be running in - prod, stage, e2e-tests, dev, etc.
+	Environment string `json:"environment,omitempty"`
```
alexeykazakov:
```suggestion
	// +optional
	Environment string `json:"environment,omitempty"`
```

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// The AuthClient contains all necessary information about the auth client
+	AuthClient AuthClient `json:"authClient,omitempty"`
```
alexeykazakov:
```suggestion
	// +optional
	AuthClient AuthClient `json:"authClient,omitempty"`
```

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+type AuthClient struct {
+	// The LibraryUrl identifies the auth library location
+	LibraryUrl string `json:"libraryUrl,omitempty"`
+
+	// The Config contains the auth config
+	Config string `json:"config,omitempty"`
+
+	// The PublicKeysUrl identifies the public keys location
+	PublicKeysUrl string `json:"publicKeysUrl,omitempty"`
```
alexeykazakov: All these fields are not optional, are they? If so then let’s remove omitempty

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
```diff
@@ -0,0 +1,83 @@
...
+// RegistrationServiceStatus defines the observed state of RegistrationService
+// +k8s:openapi-gen=true
+type RegistrationServiceStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Conditions is an array of current Registration Service deployment conditions
+	// Supported condition types:
+	// Deploying, and Deployed
```
alexeykazakov: Wouldn’t just one condition Ready with multiple reasons: Deploying, Deployed, Failing, etc, be enough and easier to handle?

alexeykazakov: Just for clarification. Let’s keep conditions as an array but IMO for now let’s use only Ready condition with multiple reasons.

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
MatousJobanek: They are optional - at least they are right now in the registration-service

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
MatousJobanek: yes, I wrote it in the wrong way - I wanted to write: "Supported condition reasons" :-)

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
MatousJobanek: stupid me - I forgot to push a newer commit with updated/fixed stuff

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
MatousJobanek: ok, now I probably understand what you mean by that. In the current registation-service, there are the default values used, so I guess that I should move the default values from registation service into the host-operator and fill it into the CR

MatousJobanek: on the other hand, wouldn't it make more sense to keep these default values inside the registration-service as they logically belong there?

MatousJobanek:
@alexeykazakov thanks for your review - I accidentally forgot to push the latest commit that changed most of the problematic parts. Anyway, I merged your suggestions and fixed the missing ones + added additional columns and it should be done now.

pkg/apis/toolchain/v1alpha1/registrationservice_types.go
alexeykazakov: Let’s keep the default values inside the reg service.
And yes the AuhClient should be optional in CRD but if it’s specified that it doesn’t make much sense to keep individual fields of the AuthClient optional. Does it make sense to you?

MatousJobanek: ok, changed in 5da71b3

MatousJobanek:
> I would keep the fields in AuthClient required (while the AuthClient itself being optional) but it’s minor.

changed in 5da71b3 

> Also we could add print column for Environment too but not sure how useful it would be...

add both image and environment cda551b 

Thanks a lot for your review :+1:

PR #65, opened 2019-12-02
feat(CRT-361): add target copying reg-service deployment to host repo

MatousJobanek:
## Description
Add a target that copies the deployment template from registration-service repo to host-operator repo - for more information see:
https://github.com/codeready-toolchain/registration-service/pull/75
https://github.com/codeready-toolchain/host-operator/pull/111

## Checks

**yes**

**yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/111

### DIFF
--- a/make/generate.mk
+++ b/make/generate.mk
@@ -9,7 +9,7 @@ MEMBER_CLUSTER_CRDS:=useraccount nstemplateset
 
 .PHONY: generate
 ## Generate deepcopy, openapi and CRD files after the API was modified
-generate: vendor generate-deepcopy generate-openapi generate-crds generate-csv
+generate: vendor generate-deepcopy generate-openapi generate-crds generate-csv copy-reg-service-deployment
 	
 .PHONY: generate-deepcopy
 generate-deepcopy:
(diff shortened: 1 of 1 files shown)


make/generate.mk
```diff
@@ -103,3 +103,7 @@ generate-kubefed-crd: vendor
 	@go install github.com/go-bindata/go-bindata/...
 	@$(GOPATH)/bin/go-bindata -pkg cluster -o ../toolchain-common/pkg/cluster/kubefedcluster_assets.go -nocompress -prefix deploy/crds/kubefed deploy/crds/kubefed
 	@rm -rf deploy/crds/kubefed
+
+.PHONY: copy-reg-service-deployment
+copy-reg-service-deployment:
+	cp ../registration-service/deploy/deployment.yaml ../host-operator/deploy/registration-service/deployment.yaml
```
xcoulon: actually, shouldn't we rename this file to `template.yaml` or `registration-service-template.yaml`. It's not just a `deployment` manifest, so the name is a bit confusing.

MatousJobanek: how about `deployment-template.yaml`? Is it fine? We already have `registration-service` in the name of the directory, so I would avoid putting it there

MatousJobanek: renamed 7e839f9

PR #80, opened 2019-12-18
feat: separate generation of deployment hack file to a script

MatousJobanek:
## Description
Creates a new script generate-deploy-hack.sh that contains logic for generation deployment hack file - the file deploys the operator bundle in a ConfigMap and creates CatalogSource that uses the ConfigMap as a source. It's usefull for testing manifests

## Checks

**yes**

**no**


scripts/generate-deploy-hack.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
+
+# Exit on error
+set -e
+
+user_help () {
+    echo "Generate CSVs and CRDs deployment hack file"
```
alexeykazakov: Can we mention here that it creates CatalogSource which uses a ConfigMap which is useful for testing CSVs / operator installations?

MatousJobanek: beter? 9babce2

alexeykazakov: Yes! Thanks.

PR #92, opened 2020-01-23
chore(CRT-427): use digest instead of tag for defining docker images

MatousJobanek:
## Description
Use digest instead of tag for defining docker images while generating CSV. The reason is the air-gapped installation of operators. see: https://docs.openshift.com/container-platform/4.2/operators/olm-restricted-networks.html

## Checks

**yes**

**no**


MatousJobanek:
this is a result of the generation of the latest commit:
https://gist.github.com/MatousJobanek/4c0500befecadbfb82dcaf27d424564c

PR #96, opened 2020-01-29
move NSTemplateSet status reasons to 'api' repo

xcoulon:
## Description
Use `NSTemplateSet` prefix
Also:
- export the label prefix so it can
be used in other pkgs
- add a `Terminating` condition to use when a resource
is terminating

updates https://issues.redhat.com/browse/CRT-405

## Checks
1. Have you run `make generate` target? **yes**

2. Does `make generate` change anything in other projects (host-operator, member-operator, toolchain-common)? **no**

### DIFF
--- a/pkg/apis/toolchain/v1alpha1/conditions.go
+++ b/pkg/apis/toolchain/v1alpha1/conditions.go
@@ -12,8 +12,6 @@ type ConditionType string
 const (
 	// ConditionReady specifies that the resource is ready
 	ConditionReady ConditionType = "Ready"
-	// ConditionTerminating specifies that the resource is being terminated/deleted (and probably deleting some related resources)
-	ConditionTerminating ConditionType = "Terminating"
 )
 
 type Condition struct {
--- a/pkg/apis/toolchain/v1alpha1/conditions.go
+++ b/pkg/apis/toolchain/v1alpha1/conditions.go
@@ -12,6 +12,8 @@ type ConditionType string
 const (
 	// ConditionReady specifies that the resource is ready
 	ConditionReady ConditionType = "Ready"
+	// ConditionTerminating specifies that the resource is being terminated/deleted (and probably deleting some related resources)
+	ConditionTerminating ConditionType = "Terminating"
 )
 
 type Condition struct {
--- a/pkg/apis/toolchain/v1alpha1/banneduser_types.go
+++ b/pkg/apis/toolchain/v1alpha1/banneduser_types.go
@@ -6,7 +6,7 @@ import (
 
 const (
 	// BannedUserEmailHashLabelKey is used for the banneduser email hash label key
-	BannedUserEmailHashLabelKey = labelKeyPrefix + "email-hash"
+	BannedUserEmailHashLabelKey = LabelKeyPrefix + "email-hash"
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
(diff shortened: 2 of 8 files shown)


pkg/apis/toolchain/v1alpha1/conditions.go
```diff
@@ -12,6 +12,8 @@ type ConditionType string
 const (
 	// ConditionReady specifies that the resource is ready
 	ConditionReady ConditionType = "Ready"
+	// ConditionTerminating specifies that the resource is being terminated/deleted (and probably deleting some related resources)
+	ConditionTerminating ConditionType = "Terminating"
```
alexeykazakov: Why do we need a separate type for it? I think using the same Ready type with a separate Terminating reason is be enough, isn't it?

tinakurian: @alexeykazakov do you think it would be a good idea to maybe have a place where we can store these consts in a more generic way? For instance, UserAccount has "Terminating" but the variable name is specific to UserAccount atm. Perhaps we could call it "TerminatingReason" for instance and store it in some place common to our controllers?

alexeykazakov: yes, we can create constants for re-used reason strings.

xcoulon: as I said on our Slack channel, there's something similar for namespaces in the `k8s.io/api/core/v1` pkg: 
```
/ These are the valid phases of a namespace.
const (
	// NamespaceActive means the namespace is available for use in the system
	NamespaceActive NamespacePhase = "Active"
	// NamespaceTerminating means the namespace is undergoing graceful termination
	NamespaceTerminating NamespacePhase = "Terminating"
)
```
(that's for the namespace "phase")

xcoulon: ok, so nervermind, I will just do
```
r.updateStatusConditions(
	nsTmplSet,
	toolchainv1alpha1.Condition{
		Type:    toolchainv1alpha1.ConditionReady,
		Status:  corev1.ConditionFalse,
		Reason:  toolchainv1alpha1.NSTemplateSetTerminatingReason,
		Message: "",
	})
```
and drop the idea of ConditionTerminating.

xcoulon:
> Looks good.
> BTW, I guess that the changes in [scripts/enrich-by-envs-from-yaml.sh](https://github.com/codeready-toolchain/api/pull/96/files#diff-ea834206f528607c04f091379f135758) were commited unintentionally ;-)

thanks for the review @MatousJobanek and yes, the change is [scripts/enrich-by-envs-from-yaml.sh](https://github.com/codeready-toolchain/api/pull/96/files#diff-ea834206f528607c04f091379f135758) is not supposed to be in this PR 🤦‍♂

xcoulon:
> > Looks good.
> > BTW, I guess that the changes in [scripts/enrich-by-envs-from-yaml.sh](https://github.com/codeready-toolchain/api/pull/96/files#diff-ea834206f528607c04f091379f135758) were commited unintentionally ;-)
> 
> thanks for the review @MatousJobanek and yes, the change is [scripts/enrich-by-envs-from-yaml.sh](https://github.com/codeready-toolchain/api/pull/96/files#diff-ea834206f528607c04f091379f135758) is not supposed to be in this PR 🤦‍♂

reverted in 4e6fc99

pkg/apis/toolchain/v1alpha1/conditions.go
tinakurian: @xcoulon sounds good, when i work on the Jira i'll work to converge the consts that are the same

PR #100, opened 2020-02-06
Refactor masteruserrecord spec

tinakurian:
## Description
Copying UserAccountSpec to masteruserrecord_types and rename to UserAccountEmbedded in order to decouple UserAccountSpec type from UserAccount. 

Relates to Jira task: https://issues.redhat.com/browse/CRT-455
Relates to Jira story: https://issues.redhat.com/browse/CRT-444

This will be done in three stages: 

1. Deprecate Embedded UA ID/Disabled
2. Change the implementation and start using the new API. Roll out to prod.
3. Delete deprecated API.

This PR is stage 1. 

## Checks
yes
yes
    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/143


pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -79,6 +79,31 @@ type MasterUserRecordStatus struct {
...
+	// Is to be used to create Identity and UserIdentityMapping resources. This field
+	// is duplicated and will be removed in the future.
+	// +optional
+	UserID string `json:"userID"`
+
+	// If set to true then the corresponding user should not be able to login
+	// "false" is assumed by default. This field is duplicated and will be
+	// removed in the future.
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
```
MatousJobanek: The reason for the new type in MUR is to avoid duplicating `UserID` and `Disabled` params

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -56,7 +56,7 @@ type UserAccountEmbedded struct {
 	SyncIndex string `json:"syncIndex"`
 
 	// The spec of the corresponding UserAccount
-	Spec UserAccountSpec `json:"spec"`
+	Spec UserAccountSpecEmbedded `json:"spec"`
```
tinakurian: @alexeykazakov is this a good name for this? or should we go with EmbeddedUserAccountSpec? I thought maybe `UserAccountSpecEmbedded` so it's similar to `UserAccountStatusEmbedded` What do you think?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
alexeykazakov: Right. But we don't want to break the API. We are planning to do in three steps:
1. Deprecate Embedded UA ID/Disabled
2. Change the implementation and start using the new API. Roll out to prod.
3. Delete deprecated API.

tinakurian: @MatousJobanek yeah, the idea is to remove them in a third PR once the changes have been made in host-operator. For now, both userID and Disabled are set to optional fields. Unless i misunderstood the task (which is possible :sweat_smile:  ) @alexeykazakov?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
alexeykazakov: +1 to UserAccountSpecEmbedded

alexeykazakov:
> I would really appreciate using composition in the combination with UserAccountSpec to make the comparison and update code much easier

@MatousJobanek what do you mean? Are you talking about extracting the common part from UserAccountSpec and UserAccountSpecEmbedded (which currently just two fields NSTemplateTier and NSTemplate) to some shared type and re-use it in both UserAccountSpec and UserAccountSpecEmbedded or something else?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
alexeykazakov: You got it right @tinakurian

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -79,6 +79,31 @@ type MasterUserRecordStatus struct {
 	UserAccounts []UserAccountStatusEmbedded `json:"userAccounts,omitempty"`
 }
 
+// UserAccountSpecEmbedded defines the desired state of UserAccount
+// +k8s:openapi-gen=true
+type UserAccountSpecEmbedded struct {
```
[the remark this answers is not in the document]
xcoulon: ah, looks like the same discussion is happening on https://github.com/codeready-toolchain/api/pull/100#pullrequestreview-354608580

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
MatousJobanek: > Right. But we don't want to break the API

it's possible to do it in one step and you shouldn't break anything if I'm not mistaken - the change will affect only one operator. But I got your point with deprecation, so I'm fine with the three steps approach

MatousJobanek:
>  Are you talking about extracting the common part from UserAccountSpec and UserAccountSpecEmbedded (which currently just two fields NSTemplateTier and NSTemplate) to some shared type and re-use it in both UserAccountSpec and UserAccountSpecEmbedded

yes, exactly. I know that it's only two fields, but it will make the comparison and assignment much easier - there won't be necessary any comparison/transformation functions

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
MatousJobanek: btw, it would be good to add this plan of doing it in three steps into the description of the PR. I know that you referenced the Jira story, but for me, the relevant info is only what is written in the GH description - Jira may become easily outdated. Thanks :-)

alexeykazakov:
> yes, exactly. I know that it's only two fields, but it will make the comparison and assignment much easier - there won't be necessary any comparison/transformation functions

Yeah.. that makes sense.

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
alexeykazakov: @MatousJobanek yes, we should add enough details of what is being done in the PR itself. +1

Regarding multiple steps... Yes, it's one operator but API and host operator leaves in different repos. So, we should either do it in multiple steps (introduce not breaking API changes, update operator, clean-up API) or create all the relevant PRs at once and merge them together when everything is done which is more messy.

Another important thing here is how we update the existing resources in the crt cluster... The existing MURs will still miss the new fields.

PR #107, opened 2020-02-24
Update to Kube 1.16.2

alexeykazakov:
## Checks
yes

yes

    - toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/81
    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/148
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/126


pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -129,6 +129,8 @@ type Cluster struct {
 // +kubebuilder:printcolumn:name="Cluster",type="string",JSONPath=`.spec.userAccounts[].targetCluster`
 // +kubebuilder:printcolumn:name="Banned",type="string",JSONPath=`.spec.banned`,priority=1
 // +kubebuilder:printcolumn:name="Disabled",type="string",JSONPath=`.spec.disabled`,priority=1
+// +kubebuilder:validation:XPreserveUnknownFields
```
xcoulon: just curious, why do we need this? I've seen the explanation on https://book.kubebuilder.io/reference/markers/crd-processing.html but I'd like to understand in which case we need this feature

alexeykazakov: For some reason NSTemplateTier CRD is broken without that annotation. It fails with an error when creating it.
Other CRDs seems to be working fine but there is an error status condition in the CRD if we don't have that annotation... Not sure why though. Maybe we are doing something wrong somewhere else.

PR #114, opened 2020-03-03
feat(CRT-478): create ChangeTierRequest that will ensure tier change

MatousJobanek:
## Description
In the detailed design doc, there is a promotion flow where MUR is changed via registration-service. I think that it's not the best approach and that we should keep the logic inside of the controllers as much as possible. Thus I created a new CRD called `ChangeTierRequest` that will trigger the change of the tier and its controller will take care of the change.

The CRD contains three params:
* `murName` - name of MUR/UserAccount the tier should be changed for
* `tierName` - name of the tier we should promote to
* `targetCluster` - cluster to identify UserAccount the tier should be changed for

and it will have one condition `complete` with only two reasons:
* `Changed` - when the change of the tier inside MUR is done
* `ChangeFailed` - when there is any problem with either loading the tier or changing the MUR

Controller:
The controller will watch only for creation of `ChangeTierRequest` resource - nothing else. When the resource is created, then the controller would:
1. load the required tier 
2. load the MUR
3. change the MUR tierName and namespaces defined in NSTemplateSet according to what is in tier
4. update MUR

## Checks

**yes**

**yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/155


MatousJobanek:
> So, if ChangeTierRequest failed we would need to create a new ChangeTierRequest for a re-try?

Technically not - if the process fails, then the controller returns an error which should trigger a new reconcile, so you could still modify the failed `ChangeTierRequest` object.

>  In Reg-service we could delete succeeded requests:

yes. I was also thinking about deleting the objects automatically, but with some delay - eg. after one or several days - we could achieve that by setting `GracePeriodSeconds` while calling the delete.

I have another idea how to avoid re-triggering the reconcile function for requests that were already processed. When the tier is changed then the controller would:
1. set the status == complete
2. add a label `changed=true` to the `ChangeTierRequest` resource

in the controller watcher, we could filter out resources containing that label. With this:
* we could watch for all changes in the `ChangeTierRequest` resources, not only for creation
* we would be sure that the reconcile won't be triggered for already processed resources. This is the main benefit as I'm not sure how it would behave after a new deployment of the operator - would it trigger the reconcile for all already existing resources when we would watch only for the creation? 
* we will be able to easily filter/select resources that are complete and ready to be deleted

alexeykazakov:
@MatousJobanek WDYT about the following:
- We create a new predicate which will accept the following events only:
  - There is no old object available (== created objects)
  - Objects with status set to "Failed".
  Or we could even make it simpler and filter out any objects with status "Complete". So, objects with empty status (== not yet handled) or failed ones will be reconciled.
- As we discussed: after successful reconcile the controller will delete the resource with `GracePeriodSeconds` set to 24 hours. So, the request will be deleted in 24 hours. If failed then update status to Failed and return an error (it will cause two re-reconciles, one because of modified status and one because we returned an error but it should not be a problem).

MatousJobanek:
> Or we could even make it simpler and filter out any objects with status "Complete". So, objects with empty status (== not yet handled) or failed ones will be reconciled.

yes, this is what I had in my mind after our discussion yesterday, so I'll try this. :+1:

PR #118, opened 2020-03-06
Make cluster resources a pointer so it's easy to use it as optional field

alexeykazakov:
## Description
Required by https://github.com/codeready-toolchain/host-operator/pull/157

## Checks
yes

no


MatousJobanek:
Looks good in general, but breaks the consistency of how we use the types in other CRDs. So let's wait for the final implementation of codeready-toolchain/host-operator#157 if it's really needed.

alexeykazakov:
> on the other hand, it's not a big deal and it will probably make the usage really simpler.

Yeah.. it's much less error prone to use a pointer to distinguish the tiers with cluster resources from the ones without them.

PR #134, opened 2020-05-01
adding notificationsent condition

tinakurian:
## Description
Adding the notificationsent condition for the notification service. 

Jira: https://issues.redhat.com/browse/CRT-511

## Checks
yes
no


pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -12,6 +12,8 @@ const (
 	MasterUserRecordUserAccountNotReady ConditionType = "UserAccountNotReady"
 	// MasterUserRecordReady means the Master User Record provisioning succeeded
 	MasterUserRecordReady ConditionType = "Ready"
+	// MasterUserRecordUserNotificationSent is used to track the status of notifications
+	MasterUserRecordUserNotificationSent ConditionType = "NotificationSent"
```
alexeykazakov, line 16: In the new design we agreed on in slack we won't know if the notification has been sent or not. We need a condition to track the notification creation status. It's up to the notification controller to handle the actual email notification. Also I would specify the type of the notification in case we need to support different notifications in the future.
So, something like "ProvisioningNotificationCreated" would be more accurate.

@sbryzak WDYT about the name of that notification type?

tinakurian: @alexeykazakov do we also want to add a reason?

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -12,6 +12,8 @@ const (
 	MasterUserRecordUserAccountNotReady ConditionType = "UserAccountNotReady"
 	// MasterUserRecordReady means the Master User Record provisioning succeeded
 	MasterUserRecordReady ConditionType = "Ready"
+	// MasterUserRecordProvisioningNotificationCreated is used to track the status of provisioned notification
+	MasterUserRecordProvisioningNotificationCreated ConditionType = "ProvisioningNotificationCreated"
```
MatousJobanek, line 16: Since you use `userprovisioned` name for naming the notification template, then I believe that you could rename it accordingly 
```suggestion
	// MasterUserRecordUserProvisionedNotificationCreated means that the Notification CR was created so the user should be notified about the successful provisioning
	MasterUserRecordUserProvisionedNotificationCreated ConditionType = "UserProvisionedNotificationCreated"
```

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
MatousJobanek: I believe that yes - for both situations:
* created: when the CR was successfuly created
* failed: when the CR creation fails

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
tinakurian: @MatousJobanek sure,sounds good - @alexeykazakov is this okay? I've updated the PR to use this name instead: ca1ba18e26acafbb409d9bfef49d6a33f923d91f

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
tinakurian: added here: ca1ba18e26acafbb409d9bfef49d6a33f923d91f

tinakurian: @sbryzak is this good?

sbryzak: > In the new design we agreed on in slack we won't know if the notification has been sent or not. We need a condition to track the notification creation status. It's up to the notification controller to handle the actual email notification. Also I would specify the type of the notification in case we need to support different notifications in the future.
> So, something like "ProvisioningNotificationCreated" would be more accurate.
> 
> @sbryzak WDYT about the name of that notification type?

I think this name is reasonable, considering that we may introduce other notification types in the future.

pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
```diff
@@ -25,6 +27,8 @@ const (
 	MasterUserRecordUnableToDeleteUserAccountsReason         = "UnableToDeleteUserAccounts"
 	MasterUserRecordUnableToRemoveFinalizerReason            = "UnableToRemoveFinalizer"
 	MasterUserRecordDisabledReason                           = disabledReason
+	MasterUserRecordUserProvisionedNotificationCreatedReason = "Created"
+	MasterUserRecordUserProvisionedNotificationFailedReason  = "Failed"
```
MatousJobanek, line 31: I believe that you could make it shorter - just 
```suggestion
	MasterUserRecordNotificationCRCreatedReason = "NotificationCRCreated"
	MasterUserRecordNotificationCRCreationFailedReason  = "NotificationCRCreationFailed"
```
or make it generic: 
```suggestion
	NotificationCRCreatedReason = "NotificationCRCreated"
	NotificationCRCreationFailedReason  = "NotificationCRCreationFailed"
```
and move it out of the masteruserrecord_types.go file

tinakurian: this looks much better! i guess for now, let's leave them here since they aren't used anywhere else, when we add the other notifications we can always move it and make it more generic. i've made the change here: c9659c3889827f95489a520d361e4c78df50d9a5

PR #136, opened 2020-05-14
feat(CRT-501): make legacy fields optional and new ones mandatory

MatousJobanek:
## Description
Since we are moving to the `templateRef` field this PR is making it mandatory and the other fields (type, revision) optional.
In addition, there is a new label `/templateref` introduced and the old ones `/tier` and `/revision` removed, since they were just duplication of the information contained in the `/templateref` and there was no usage of it

## Checks

**yes**

**yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/205
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/#

### DIFF
--- a/pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
+++ b/pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
@@ -7,12 +7,11 @@ import (
 )
 
 const (
-	OwnerLabelKey      = LabelKeyPrefix + "owner"
-	RevisionLabelKey   = LabelKeyPrefix + "revision"
-	TypeLabelKey       = LabelKeyPrefix + "type"
-	TierLabelKey       = LabelKeyPrefix + "tier"
-	ProviderLabelKey   = LabelKeyPrefix + "provider"
-	ProviderLabelValue = "codeready-toolchain"
+	OwnerLabelKey       = LabelKeyPrefix + "owner"
+	TypeLabelKey        = LabelKeyPrefix + "type"
+	TemplateRefLabelKey = LabelKeyPrefix + "templateref"
+	ProviderLabelKey    = LabelKeyPrefix + "provider"
+	ProviderLabelValue  = "codeready-toolchain"
 )
 
 // These are valid status condition reasons of a NSTemplateSet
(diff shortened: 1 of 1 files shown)


pkg/apis/toolchain/v1alpha1/nstemplateset_types.go
```diff
@@ -7,12 +7,11 @@ import (
 )
 
 const (
-	OwnerLabelKey      = LabelKeyPrefix + "owner"
-	RevisionLabelKey   = LabelKeyPrefix + "revision"
-	TypeLabelKey       = LabelKeyPrefix + "type"
-	TierLabelKey       = LabelKeyPrefix + "tier"
```
xcoulon, line 13: why did you remove `TierLabelKey`?

MatousJobanek: I decided to remove the `/tier` label as well as the `/revision` label because they were redundant. All information that is needed is provided by either `/templateref` label or by the `/type` label. 
The reason for the removal was that having multiple (redundant) labels in place makes the stuff a bit more complicated - we would need to keep all of them in sync, but there would be a question which of the labels should be taken as the main one for the comparisons. All of them or only one of them. With my proposal, we use only the `/templateref` label.

I talked about it with @alexeykazakov and he agreed on dropping them, but if you see that it's a must to have them there, I can bring it back, but I'm not sure if it brings us any benefit. Information about the tier is stored in the NSTemplateSet which should be enough for admin purposes.

MatousJobanek: ok, maybe at some time the `/tier` could be useful and if we don't use the label for comparison, then it shouldn't complicate the code much. Let me bring it back

MatousJobanek: done in 4f5f9cc
and related commit in member-operator PR: https://github.com/codeready-toolchain/member-operator/pull/156/commits/a6ae1dc9601dda761b97363ae81109f900710443

xcoulon: thanks for taking my comment into account, @MatousJobanek 🙌🙏

PR #145, opened 2020-06-05
feat: support both versions of operator-sdk 0.16.0 and 0.17.1

MatousJobanek:
## Description

The changes in this PR brings support of operator-sdk to our scripts. With these changes, we can use both versions of operator-sdk 0.16.0 and 0.17.1

This is very first step of supporting operator-sdk 0.17.1. After this, we will
1. migrate all repositories to use operator-sdk 0.17.1 in their dockerfiles
2. drop support of operator-sdk 0.16.0 in our scripts
3. adapt the structure of the repositories and the logic of our scripts to properly use operator-sdk 0.17.1
4. ... and some more steps that are still not clear right now :-) ...

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -7,7 +7,7 @@ image:https://codecov.io/gh/codeready-toolchain/api/branch/master/graph/badge.sv
 Requires:
 
 * Go version 1.13 - download for your development environment https://golang.org/dl/[here].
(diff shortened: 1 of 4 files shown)


xcoulon:
> LGTM and works on my machine(tm) ;)

could you also mention support for version 0.17.1 in https://github.com/codeready-toolchain/api/blob/master/README.adoc?

MatousJobanek:
README changed in 94f0fd4

PR #156, opened 2020-06-16
notification delivered reason

tinakurian:
## Description
This PR adds a delivered reason to the notification cr which can be used when setting the notification.condition. reason field.

Related Jira: https://issues.redhat.com/browse/CRT-508
Related PR: 

## Checks
yes
no


pkg/apis/toolchain/v1alpha1/notification_types.go
```diff
@@ -9,6 +9,9 @@ const (
 
 	// NotificationDelivered reflects whether the notification has been delivered to the user
 	NotificationDelivered ConditionType = "Delivered"
+
+	// Status condition reasons
+	NotificationServiceDeliveredReason = "Delivered"
```
xcoulon, line 14: it's a bit weird to have `Delivered` for both the condition type and the reason. At the end of the day, the Notification condition will be something like:
```
Type:    Delivered
Status:  true
Reason:  Delivered
```

I wonder if having `Type: "Complete"` as we do in https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/changetierrequest/changetierrequest_controller.go#L202 would be better? WDYT?

alexeykazakov:
I just realized that Delivered is not quite accurate even as the type name
:)
We don’t know if the notification (email) was delivered or not.
The correct status would be Sent.
“Complete” reason looks a bit confusing to me. What does “Complete” mean?
I think it’s ok to have both the type and reason to be Sent.

pkg/apis/toolchain/v1alpha1/notification_types.go
tinakurian: sounds good! changed to "sent" as Alexey had suggested. Thanks for pointing this out Xaiver!

PR #157, opened 2020-06-17
Adding error condition

tinakurian:
## Description
This is to address the PR comment found here: https://github.com/codeready-toolchain/host-operator/pull/226#discussion_r441206596

## Checks
yes

no

Related PRs: coming soon..


pkg/apis/toolchain/v1alpha1/conditions.go
```diff
@@ -19,6 +19,9 @@ const (
 	disabledReason     = "Disabled"
 	terminatingReason  = "Terminating"
 	updatingReason     = "Updating"
+
+	// Condition types
+	deletedCondition = "Deleted"
```
alexeykazakov, line 24: Well... we are not interested in tracking the deleted resources. When they are deleted they are already gone. There is nothing to track in the status. But tracking errors when we can't delete is what we would want.
So, I would rather introduce DeletionError condition type. Which is set to "true" when we have an error during deletion.

@MatousJobanek any better ideas?

tinakurian: no worries, i'll change it if DeletionError is more clear for you

tinakurian: changed here: a44edc43c83d31f4be9f6d2e3505736309c4d9cd

PR #158, opened 2020-06-18
Add member status resource

rajivnathan:
### Add `MemberStatus` resource
Related Issue: https://issues.redhat.com/browse/CRT-538

As a developer, I want to store toolchain status information within the system so that it can be retrieved easily and used to inform administrators about the status of the toolchain.

MemberStatus:

```
kind: MemberStatus
metadata:
 name: toolchain-member-status
spec: {}
status: 
  memberOperator:
    revision: 9b8caf21ce2115c376667bb0071f5bde9232b961
    buildTime: "2020-05-20T15:09:31Z"
    deployment:
      name: member-operator 
    conditions:
    - type: Ready
      status: False
      reason: MemberOperatorDeploymentFailing
      message: Can’t pull the image…
      lastUpdateTime: "2020-05-20T15:09:31Z"
      lastTransitionTime: "2020-05-20T15:09:31Z"
  hostConnection:
    conditions:
    - type: Ready
      status: False
      reason: KubefedNotReady
      message: Cannot connect to cluster
      lastUpdateTime: "2020-05-20T15:09:31Z"
      lastTransitionTime: "2020-05-20T15:09:31Z"
  conditions:
  - type: Ready
    status: False
    reason: MemberOperatorDeploymentFailing
    message: Can’t pull the image…
    lastUpdateTime: "2020-05-20T15:09:31Z"
    lastTransitionTime: "2020-05-20T15:09:31Z"
``` 

### Add `LastUpdatedTime` condition field

`LastUpdatedTime` is being added for `ToolchainStatus` and `MemberStatus` status conditions. `LastTransitionTime` indicates the last time a particular condition was changed. However, ToolchainStatus and MemberStatus resources are updated using the controller reconciler requeue method which means it is updated on a consistent interval.

The status itself may not change from one update to the next so it would not be accurate to use `LastTransitionTime` to indicate the updated status. It is also not sufficient to rely on `LastTransitionTime` alone because if the controller fails then no updates will be processed, which means the status may not be correct. The user requires `LastUpdatedTime` to indicate the last time the status was checked. If it was not checked for a long period of time then it could indicate a problem to the user.

## Checks
yes

yes

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/229
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/173
    - toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/112


pkg/apis/toolchain/v1alpha1/memberstatus_types.go
```diff
@@ -0,0 +1,111 @@
...
+
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// MemberStatusSpec defines the desired state of MemberStatus
+// +k8s:openapi-gen=true
+type MemberStatusSpec struct {
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// spec is intentionally empty since only the status fields will be used for reporting status of the toolchain
```
alexeykazakov, line 17: I was thinking that we could have a timeout between status updates in the spec. But that's OK to keep it as hardcoded value for now.

alexeykazakov:
Can you please update the links to the corresponding host/member PR in the description?

pkg/apis/toolchain/v1alpha1/memberstatus_types.go
rajivnathan: Ah, that's a good idea. We could also have it as config I guess?

alexeykazakov: We could have it as config but the config would be used to set the Spec when we are creating Status resources.

rajivnathan:
> Can you please update the links to the corresponding host/member PR in the description?

Sorry, updated it now

PR #162, opened 2020-06-26
fix: delete bundle files when doing final releases

MatousJobanek:
## Description
When the scripts generate a new release manifests (in the old bundle format), then it should not include the `bundle.Dockerfile` nor the `metadata/`


scripts/create-release-bundle.sh
```diff
@@ -54,10 +54,14 @@ copy_manifests_to_versioned_dir_and_adjust_package_file
 # delete the default bundle directory that is used as a template
 rm -rf ${BUNDLE_DIR}
 
-# copy everything from package dir to manifests
+# make sure that manifests dire at the root of the project exists
```
alexeykazakov, line 57:
```suggestion
# make sure that a manifests dir exists at the root of the project
```

PR #163, opened 2020-06-29
Add required reason constants for Notification Controller

sbryzak:
## Description
This PR just adds a couple of new reason constants.

Fixes https://issues.redhat.com/browse/CRT-678

**NotificationContextError** is set when there is an error while attempting to create a new notification context.
**NotificationDeliveryError** is set when there is an error during delivery of a notification.

## Checks
1. Have you run `make generate` target? **[no]**

2. Does `make generate` change anything in other projects (host-operator, member-operator, toolchain-common)? **[no]** 

  *Not applicable - this PR only adds new constants*

### DIFF
--- a/pkg/apis/toolchain/v1alpha1/notification_types.go
+++ b/pkg/apis/toolchain/v1alpha1/notification_types.go
@@ -17,7 +17,7 @@ const (
 	NotificationSentReason          = "Sent"
 	NotificationDeletionErrorReason = "UnableToDeleteNotification"
 	NotificationContextErrorReason  = "NotificationContextError"
-	NotificationDeliveryErrorReason = "NotificationDeliveryError"
+	NotificationDeliveryErrorReason = "DeliveryError"
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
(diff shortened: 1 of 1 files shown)


pkg/apis/toolchain/v1alpha1/notification_types.go
```diff
@@ -16,6 +16,8 @@ const (
 	// Status condition reasons
 	NotificationSentReason          = "Sent"
 	NotificationDeletionErrorReason = "UnableToDeleteNotification"
+	NotificationContextErrorReason  = "NotificationContextError"
+	NotificationDeliveryErrorReason = "NotificationDeliveryError"
```
MatousJobanek, line 20: I'm curious - do we really need to have the word "notification" as part of the reason? It is the status of the Notification CR, I'd say that it's logical that the errors relate to the notification
```suggestion
	NotificationContextErrorReason  = "ContextError"
	NotificationDeliveryErrorReason = "DeliveryError"
```

sbryzak: "NotificationContext" is an actual type, so for this one it's necessary.  For the delivery error though it's not required - I actually like "DeliveryError" as an alternative, so I'll change it.  Thanks :)

PR #165, opened 2020-07-08
chore: update readme with a new operator bundle format changes

MatousJobanek:
## Description
* Updates the README in scripts directory to reflect the changes related to the newly added support of the new operator bundle format
* drops  `push-to-quay-nightly.sh` since it's not used anywhere


scripts/push-bundle-and-index-image.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 additional_help() {
-    echo "Important info: push-to-quay-nightly.sh scripts overrides all the parameters but \"--project-root\", \"--embedded-repo\", \"--quay-namespace\" and \"--operator-name\", so use only these to specify necessary values."
+    echo "Important info: push-to-quay-nightly.sh scripts overrides all the parameters but \"--project-root\", \"--embedded-repo\", \"--quay-namespace\", \"--index-image\", \"--image-builder\" and \"--operator-name\", so use only these to specify necessary values."
```
alexeykazakov, line 4: You are removing `push-to-quay-nightly.sh` so I guess this should be modified.

MatousJobanek: yeah, this is the result of copy-paste driven development without changing it :-) 
Good catch, thanks :+1:

MatousJobanek: fixed in 2626555

PR #167, opened 2020-07-08
Idler API for managing user's pods

alexeykazakov:
This PR introduce a new `Idler` resource which enables pod "idling" 

Also see:
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/177


pkg/apis/toolchain/v1alpha1/idler_types.go
```diff
@@ -0,0 +1,85 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// These are valid conditions of an Idler
+const (
+	// IdlerReady means the Idler is in a healthy state
+	IdlerReady ConditionType = "Ready"
```
MatousJobanek, line 8: Why don't we use the generic ConditionType "Ready" that is in the conditions.go - we use the same conditions
https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/conditions.go#L13-L14

pkg/apis/toolchain/v1alpha1/idler_types.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Idler enables automatic idling of payloads in a user namespaces
+// where the name of the Idler matches the name of the corresponding namespace.
+// For example an Idler with "foo" name will be managing pods in namespace "foo".
+// +k8s:openapi-gen=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
```
MatousJobanek, line 63: is it cluster scoped to avoid users to modify the CR, or is there any other reason? I'm asking since it looks like a perfect case for a namespaced scoped resource. I know that we had discussion about that, but my "gold fish memory" forgot the reasoning ¯\_(ツ)_/¯

pkg/apis/toolchain/v1alpha1/idler_types.go
```diff
@@ -0,0 +1,85 @@
...
+// +kubebuilder:validation:XPreserveUnknownFields
```
MatousJobanek, line 64: How about showing the timeout and info from the status condition:
```suggestion
// +kubebuilder:printcolumn:name="Timeout",type="integer",JSONPath=`.spec.timeoutSeconds`
// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].status`
// +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
// +kubebuilder:validation:XPreserveUnknownFields
```

pkg/apis/toolchain/v1alpha1/idler_types.go
alexeykazakov: Yep. My first implementation was with namespaced Idlers. But because of the permission issues (users with "edit" role will be able to edit Idlers in their namespaces) I had to switch it to cluster scope.

pkg/apis/toolchain/v1alpha1/idler_types.go
alexeykazakov: Copy-paste from MUR ;)
Will switch to ConditiionReady.

pkg/apis/toolchain/v1alpha1/idler_types.go
alexeykazakov: +1

pkg/apis/toolchain/v1alpha1/idler_types.go
alexeykazakov: Done in https://github.com/codeready-toolchain/api/pull/167/commits/7a94abba3b8eae5269d248b4430093ce5b277a42

alexeykazakov:
> As for the scope - I understand the reason. I'm still thinking if we shouldn't rather keep the Namespaced scope and fix the issue with the user changes of the Idler CRs by watching the inner namespace resources and changing it back when the user modifies it manually. But I understand that this is not implemented yet and it's a potential risk.
We can always try to change it later although changing the scope will be pretty hard

There is multiple potential solution for that:

1. IMO the deal one would for OLM to implement a new feature to mark some CRDs as truly "internal" so they are not added to the aggregated admin/edit/view roles. I've asked about that in the operators-framework mailing list and while they agreed that it could be potentially a useful feature we can't expect it to be implemented anytime soon (but I think we could try to contribute this feature to OLM).
2. Create an admission webhook which would reject all delete and update requests from not authorized accounts. But creating a webhook and maintaining it can be pretty painful because of the certificate management. The latest operator-sdk actually introduced `operator-sdk create webhook` which is an integration with kube-builder and which helps a lot with webhooks. But the webhook it creates is a special operator-sdk/kube-builder interface implementation and can't be used for RBAC (there is no information about who/what sent the request).
3. As you mentioned we could just re-create/restore the modified/deleted Idlers by our NSTemplateSet controller. And I think we should do it anyway (to restore all resources it created). But it requires additional work and as you also mentioned let's consider it later.

PR #168, opened 2020-07-10
Add toolchain status

rajivnathan:
### Add `ToolchainStatus` resource
Related Issue: https://issues.redhat.com/browse/CRT-673
This is a follow-up to the memberstatus PR: https://github.com/codeready-toolchain/api/pull/158

As a developer, I want to store toolchain status information within the system so that it can be retrieved easily and used to inform administrators about the status of the toolchain.

This PR adds the `ToolchainStatus` resource which will apply to the host cluster. The `MemberStatus` resource is also updated to make `MemberOperator` and `HostConnection` optional fields because when the memberstatus is included as part of the toolchainstatus, only the overall status (ie. MemberStatus.status.Conditions) should be shown to avoid having too much information in the `ToolchainStatus` thereby making it difficult to read.

`ToolchainStatus` example:
```
kind: ToolchainStatus
metadata:
  annotations:
    toolchain.dev.openshift.com/last-applied-configuration: '{"metadata":{"name":"toolchain-status","namespace":"rsenthil-host-operator","creationTimestamp":null},"spec":{},"status":{"hostOperator":{"version":"","revision":"","buildTimestamp":"","deploymentName":""},"registrationService":{"deployment":{"name":""},"registrationServiceResources":{},"health":{"alive":"","buildTime":"","environment":"","revision":"","startTime":""}}}}'
  creationTimestamp: "2020-07-09T19:06:04Z"
  generation: 1
  name: toolchain-status
  namespace: rsenthil-host-operator
  resourceVersion: "2422479"
  selfLink: /apis/toolchain.dev.openshift.com/v1alpha1/namespaces/rsenthil-host-operator/toolchainstatuses/toolchain-status
  uid: 717bedf3-f15b-460a-98a8-a49898c8253b
spec: {}
status:
  conditions:
  - lastTransitionTime: "2020-07-09T19:09:00Z"
    lastUpdatedTime: "2020-07-09T19:09:00Z"
    reason: AllComponentsReady
    status: "True"
    type: Ready
  hostOperator:
    buildTimestamp: "2020-07-09T19:04:11Z"
    conditions:
    - lastTransitionTime: "2020-07-09T19:09:00Z"
      lastUpdatedTime: "2020-07-09T19:09:00Z"
      reason: DeploymentReady
      status: "True"
      type: Ready
    deploymentName: host-operator
    revision: 7917eb39624707d30c971950ebe147e7d584558a-dirty
    version: 0.0.1
  members:
  - clusterName: member-crc-m27h4
    memberStatus:
      conditions:
      - lastTransitionTime: "2020-07-09T19:08:57Z"
        lastUpdatedTime: "2020-07-09T19:08:57Z"
        reason: AllComponentsReady
        status: "True"
        type: Ready
  registrationService:
    deployment:
      conditions:
      - lastTransitionTime: "2020-07-09T19:09:00Z"
        lastUpdatedTime: "2020-07-09T19:09:00Z"
        reason: DeploymentReady
        status: "True"
        type: Ready
      name: registration-service
    health:
      alive: "true"
      buildTime: "0"
      conditions:
      - lastTransitionTime: "2020-07-09T19:09:00Z"
        lastUpdatedTime: "2020-07-09T19:09:00Z"
        reason: RegServiceReady
        status: "True"
        type: Ready
      environment: dev
      revision: 64af1be5c6011fae5497a7c35e2a986d633b3421-dirty
      startTime: "2020-07-09T19:06:15Z"
    registrationServiceResources:
      conditions:
      - lastTransitionTime: "2020-07-09T19:06:06Z"
        reason: Deployed
        status: "True"
        type: Ready
```

## Checks
yes

yes

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/241
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/178
    - toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/115


pkg/apis/toolchain/v1alpha1/conditions.go
```diff
@@ -24,6 +24,32 @@ const (
 	deletionError = "DeletionError"
 )
 
+// These are valid status condition reasons for Toolchain status
+const (
+	// overall status condition reasons
+	ToolchainStatusReasonAllComponentsReady = "AllComponentsReady"
```
alexeykazakov, line 30: We keep all condition types and reasons in the corresponding *_type.go files. Not in conditions.go.
Let's do the same for the status types for the sake of consistency.

rajivnathan: Done, moved it to toolchainstatus_types.go

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -0,0 +1,201 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// These are valid status condition reasons for Toolchain status
+const (
+	// overall status condition reasons
+	ToolchainStatusReasonAllComponentsReady = "AllComponentsReady"
```
alexeykazakov, line 10: We try to follow the following format of the reasons:
`<Kind><ReasonString>Reason = "<ReasonString>"`
For example:
`UserAccountUnableToCreateIdentityReason = "UnableToCreateIdentity"`
For this particular reason it should be:
```suggestion
	ToolchainStatusAllComponentsReadyReason = "AllComponentsReady"
```

Lets follow this rule for the sake of constancy.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -0,0 +1,201 @@
...
+// +k8s:openapi-gen=true
+type ToolchainStatusStatus struct {
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// HostOperator is the status of a toolchain host operator
+	HostOperator HostOperatorStatus `json:"hostOperator"`
+
+	// RegistrationService is the status of the registration service
+	RegistrationService HostRegistrationServiceStatus `json:"registrationService"`
```
alexeykazakov, line 54: Are we OK to keep have them as required fields? It means if there is Status then it must always have something in "hostOperator" and "registrationService".

rajivnathan: At the moment we are okay to keep them as required fields because we add conditions even when there are errors so there should be something in them in all cases. But if we think we'd want to make them optional in the future then it may be better to make them optional now so we don't need to deal with api breakages.

alexeykazakov: Maybe having them optional will be less error prone. Even if the current controller implementation works fine with required fields we can accidentally change it in the future and we have limited ability to test errors in e2e tests so it's easy to miss this.
Basically I would keep everything in Status sub-resource as optional. Just in case.

rajivnathan:
> Also, shouldn't we do this change it two steps?

>1. Update MemberStatus API + Member Status controller (in member operator)
>2. When 1 merged and deployed: Introduce ToolchainStatus API & Controller which uses the updated MemberStatus.

>In this case we can guarantee that Host Operator works well with both member api: the current one & new one.

I actually tried the changes locally without any changes to the Host Operator and the memberstatus code works without errors. That explains why the member-operator PR builds and e2e tests are green: https://github.com/codeready-toolchain/member-operator/pull/178

So if that's correct then I don't think the host operator needs to work with the existing memberstatus api because it does not have the toolchainstatus controller yet? The host-operator PR https://github.com/codeready-toolchain/host-operator/pull/241 adds the toolchainstatus controller so it will only start using the memberstatus after that PR is merged. However, doing it in the two steps you mentioned is still cleaner and I can do it that way if it's preferred.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
rajivnathan: Ah, ok makes sense. I've updated them.

alexeykazakov:
@rajivnathan OK. We can keep it as is.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -48,10 +48,12 @@ type ToolchainStatusStatus struct {
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 
 	// HostOperator is the status of a toolchain host operator
-	HostOperator HostOperatorStatus `json:"hostOperator"`
+	// +optional
+	HostOperator *HostOperatorStatus `json:"hostOperator"`
```
alexeykazakov, line 52:
```suggestion
	HostOperator *HostOperatorStatus `json:"hostOperator,omitempty"`
```

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -48,10 +48,12 @@ type ToolchainStatusStatus struct {
...
 
 	// RegistrationService is the status of the registration service
-	RegistrationService HostRegistrationServiceStatus `json:"registrationService"`
+	// +optional
+	RegistrationService *HostRegistrationServiceStatus `json:"registrationService"`
```
alexeykazakov, line 56:
```suggestion
	RegistrationService *HostRegistrationServiceStatus `json:"registrationService,omitempty"`
```

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
rajivnathan: c401261cd16ca217a834e89a6008c35dc7115956

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
rajivnathan: c401261cd16ca217a834e89a6008c35dc7115956

PR #172, opened 2020-07-29
feat: introduce ToolchainCluster CRD that will replace KubeFedCluster

MatousJobanek:
## Description
This PR only introduces ToolchainCluster CRD that will replace the usage of KubeFedCluster CRD in the next PRs

## Checks

**yes**

**yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/256
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/184


pkg/apis/toolchain/v1alpha1/toolchaincluster_types.go
```diff
@@ -0,0 +1,152 @@
+package v1alpha1
+
+/*
+Copyright 2018 The Kubernetes Authors.
```
alexeykazakov, line 4: I guess we should remove this.

pkg/apis/toolchain/v1alpha1/toolchaincluster_types.go
```diff
@@ -0,0 +1,152 @@
...
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ToolchainClusterConditionType string
+
+// These are valid conditions of a cluster.
+const (
+	// ClusterReady means the cluster is ready to accept workloads.
+	ToolchainClusterReady ToolchainClusterConditionType = "Ready"
```
tinakurian, line 31: this is minor; i think the standard for golang comments here is to start the comment with "ToolchainClusterReady..."

pkg/apis/toolchain/v1alpha1/toolchaincluster_types.go
```diff
@@ -0,0 +1,152 @@
...
+	// ClusterOffline means the cluster is temporarily down or not reachable
+	ToolchainClusterOffline ToolchainClusterConditionType = "Offline"
```
tinakurian, line 33: same here

pkg/apis/toolchain/v1alpha1/toolchaincluster_types.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ToolchainClusterConditionType string
+
+// These are valid conditions of a cluster.
+const (
+	// ClusterReady means the cluster is ready to accept workloads.
```
tinakurian, line 30:
```suggestion
	// ToolchainClusterReady means the cluster is ready to accept workloads.
```

pkg/apis/toolchain/v1alpha1/toolchaincluster_types.go
```diff
@@ -0,0 +1,152 @@
...
+	ToolchainClusterReady ToolchainClusterConditionType = "Ready"
+	// ClusterOffline means the cluster is temporarily down or not reachable
```
tinakurian, line 32:
```suggestion
	// ToolchainClusterOffline means the cluster is temporarily down or not reachable
```

pkg/apis/toolchain/v1alpha1/toolchaincluster_types.go
MatousJobanek: wrong copy-paste :grimacing: 
good catch, thanks

pkg/apis/toolchain/v1alpha1/toolchaincluster_types.go
MatousJobanek: you are right. I'm always confused when I have to add a copy of the licence clause and when not.
removed c2f020d

alexeykazakov: Keeping the licence itself is fine. I was referring to the Kube Copyright actually. But it's fine now. Our repo does explicitly include the licence. So, I guess we don't have to duplicate it here anyway.

PR #178, opened 2020-08-19
Add deactivation days field to NSTemplateTier

rajivnathan:
## Description
Adds a new _optional_ field `DeactivationTimeoutDays` to NSTemplateTier that will be used to determine when users should be automatically deactivated. The value is an integer that represents the total number of days the user's account will be active and is configured at the tier level.

## Checks
yes

yes

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/265

### DIFF
--- a/pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go
+++ b/pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go
@@ -27,6 +27,9 @@ type NSTemplateTierSpec struct {
 	// the cluster resources template (for cluster-wide quotas, etc.)
 	// +optional
 	ClusterResources *NSTemplateTierClusterResources `json:"clusterResources,omitempty"`
+
+	// +optional
+	DeactivationTimeoutDays *int `json:"deactivationTimeoutDays,omitempty"`
 }
 
 // NSTemplateTierNamespace the namespace definition in an NSTemplateTier resource
--- a/pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go
+++ b/pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go
@@ -28,8 +28,9 @@ type NSTemplateTierSpec struct {
 	// +optional
 	ClusterResources *NSTemplateTierClusterResources `json:"clusterResources,omitempty"`
 
+	// the period (in days) after which users within the tier will be deactivated
 	// +optional
-	DeactivationTimeoutDays *int `json:"deactivationTimeoutDays,omitempty"`
+	DeactivationTimeoutDays int `json:"deactivationTimeoutDays,omitempty"`
 }
 
 // NSTemplateTierNamespace the namespace definition in an NSTemplateTier resource
--- a/pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
+++ b/pkg/apis/toolchain/v1alpha1/masteruserrecord_types.go
@@ -31,7 +31,6 @@ const (
 	MasterUserRecordNotificationCRCreatedReason              = "NotificationCRCreated"
 	MasterUserRecordNotificationCRCreationFailedReason       = "NotificationCRCreationFailed"
 	MasterUserRecordUserIDLabelKey                           = LabelKeyPrefix + "user-id"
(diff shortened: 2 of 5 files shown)


pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go
```diff
@@ -27,6 +27,9 @@ type NSTemplateTierSpec struct {
 	// the cluster resources template (for cluster-wide quotas, etc.)
 	// +optional
 	ClusterResources *NSTemplateTierClusterResources `json:"clusterResources,omitempty"`
+
+	// +optional
+	DeactivationTimeoutDays *int `json:"deactivationTimeoutDays,omitempty"`
```
MatousJobanek, line 32: do we really need to use the pointer when it's a basic type?

MatousJobanek (commented):
How do you want to configure the exclusion regex? I thought that it should have been in the NsTemplateTier, but if you want to pass it as a param through an env config file or ConfigMap, then it is fine too.

pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go
rajivnathan: Unfortunately, I believe yes. We don't want to introduce a breaking change to the API since it's an existing resource and as I found out the hard way in the past until @xcoulon informed me optional fields need to be pointers, it's the only way to make it truly optional. Otherwise, there would be errors if resources are created without that field. @alexeykazakov mentioned as a best practice when introducing new fields to existing resources we should make them optional then update existing resources then we can make the new field required if we want.

Example error of when I introduced lastUpdatedTime:
```
{"level":"error","ts":1592830271.30579,"logger":"controller-runtime.controller","msg":"Reconciler error","controller":"masteruserrecord-controller","request":"xcoulon-host-22144423/cheesecakelover00","error":"update of the MasterUserRecord failed while synchronizing with UserAccount status from the cluster 'member-xcoulon-20200622-zw87h': MasterUserRecord.toolchain.dev.openshift.com \"cheesecakelover00\" is invalid: status.userAccounts.conditions.lastUpdatedTime: Invalid value: \"null\": status.userAccounts.conditions.lastUpdatedTime in body must be of type string: \"null\"","errorVerbose":"MasterUserRecord.toolchain.dev.openshift.com \"cheesecakelover00\" is invalid: status.userAccounts.conditions.lastUpdatedTime: Invalid value: \"null\": status.userAccounts.conditions.lastUpdatedTime in body must be of type string: \"null\"\nupdate of the MasterUserRecord failed while synchronizing with UserAccount status from the cluster 'member-xcoulon-20200622-zw87h'
...
```

rajivnathan:
I think there was some discussion about the exclusion list being configured at the global level, not at the tier level so that's why I didn't include it in NSTemplateTier now. If we want it to be in the spec then I can put up a separate PR for that, if we want it to be global configuration then including it in a ConfigMap would work.

pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go
alexeykazakov: That's right. But with the basic types (like `int`) it might work as a truly optional field without using a pointer. We do need to use pointers for our own types (arbitrary structs) but it might be OK to use just `int`. Can you please try that?
- Create some CR for that old (without that field) CRD in your cluster
- Update the CRD so it now have that new optional field (but no pointer)
- See if there is any validation errors with the existing CRs.

MatousJobanek:
> I think there was some discussion about the exclusion list being configured at the global level, not at the tier level so that's why I didn't include it in NSTemplateTier now. If we want it to be in the spec then I can put up a separate PR for that, if we want it to be global configuration then including it in a ConfigMap would work.

Yup :+1: 
Having it as a global configuration (probably in the env config files) is the appropriate solution from my point of view. I wasn't just sure what was the final decision on this in the doc - that's why I asked.  :-)

pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go
rajivnathan: I just tried it out and it looks like optional primitive types without pointers are okay like you said. Did the same test with a non-primitive and it crashed the host-operator pod. So I'll make the change. Thanks both of you!

PR #181, opened 2020-08-24
chore: upgrade operator-sdk to 0.19.2 and go to 1.14.7

MatousJobanek:
## Description
upgrade operator-sdk to 0.19.2 and go to 1.14.7
it also bumps versions of the Kube dependencies to v0.18.3


make/generate.mk
```diff
@@ -72,23 +72,23 @@ endif
 .PHONY: generate-crds
 generate-crds: vendor prepare-host-operator prepare-member-operator
 	@echo "Re-generating the Toolchain CRD files..."
-	$(Q)go run $(shell pwd)/vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go crd:trivialVersions=true \
-	paths=./pkg/apis/... output:crd:dir=deploy/crds output:stdout
+	rm -rf deploy/crds/ 2>/dev/null || true
+	operator-sdk generate crds
 	@echo "Dispatching CRD files in the 'host-operator' and 'member-operator' repositories..."
     # When dispatching CRD files we delete two first lines of CRDs ("\n----\n") to make a single manifest file out of the original multiple manifest file
     # Also we remove the line with 'type: object' from validation.openAPIV3Schema.properties path because it's incompatible with kube 1.11 which is used by minishift
```
alexeykazakov, line 79: These two comments are obsolete now. We do not need to do this hack anymore if we use `operator-sdk generate crds` :)

xcoulon:
@MatousJobanek none of your PRs related to this Operator SDK upgrade seem update to the new layout as explained in https://v0-19-x.sdk.operatorframework.io/docs/golang/project_migration_guide/ 

Is it something you intend to address later?

make/generate.mk
MatousJobanek: fixed in 92bd9c0

MatousJobanek:
> Is it something you intend to address later?

@xcoulon Yes. Let's postpone this work because it will require significant changes in our repos as well as in our scripts. I'm afraid that it will take a few days to finish it so I'd like to make the migration in the next milestone. 
Since 0.19.x still supports the old layout and "only deprecates it" then I believe that we should be fine for now.

MatousJobanek:
btw. Also in the doc that you referenced, there is a note:
> Note It is recommended that you have your project upgraded to the latest SDK release version before following the steps of this guide to migrate to Kubebuilder layout.

So, I believe that we should be fine with our approach :-)

xcoulon:
> btw. Also in the doc that you referenced, there is a note:
> 
> > Note It is recommended that you have your project upgraded to the latest SDK release version before following the steps of this guide to migrate to Kubebuilder layout.
> 
> So, I believe that we should be fine with our approach :-)

ok, thanks for your response and clarification. I wanted to make sure we would migrate to the new layout at some point, otherwise I believe we'll run into issues in the future. But I guess we can postpone this change to the next SDK migration (ie, when using 1.0.0)

PR #182, opened 2020-08-31
feat: add counts to ToolchainStatus

MatousJobanek:
## Description
Adds MUR and UserAccount counts to ToolchainStatus.
it will contain:
* number of MasterUserRecords in host cluster
* number of UserAccounts in member cluster (mapped by cluster name)

This will be then used by Automatic approval logic

this is first step for CRT-788

## Checks

yes

yes

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/272


pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -72,6 +72,15 @@ type ToolchainStatusStatus struct {
...
 	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// Number of MasterUserRecords created within the host cluster
+	// +optional
+	MasterUserRecordCount int `json:"masterUserRecordCount,omitempty"`
+
+	// A map of number of UserAccounts per cluster created within the member clusters
+	// +optional
+	// +patchStrategy=merge
+	UserAccountsPerClusterCounts map[string]int `json:"userAccountsPerClusterCounts,omitempty" patchStrategy:"merge"`
```
rajivnathan, line 83: ToolchainStatus currently only has status fields pertaining directly to the host cluster except for the `Members` field, that's where all the member-related status is currently. Should this be added to the `MemberStatusStatus` instead? The information would still be retrievable from ToolchainStatus but it would be organized under the `Members` field.

rajivnathan: Perhaps I should have made `Members` a map 🤔

alexeykazakov: Yeah I wonder the same. Wouldn't moving the count to the Members be better?
I think we don't have to make it a map though. Map would be needed if we keep the counts in the root of the spec. But Members still can be a list. Each Member has a name anyway.

MatousJobanek: I cannot put it into the `MemberStatusStatus` part because it is part of the `MemberStatus` CRD which lives in member cluster, however, the counting is done only in host cluster. That's why I added it as a separated field - the number is not collected from the member cluster and the respective `MemberStatus` CRs but is counted in host.

Anyway, I see your point. Originally I was thinking about creating a subsection `counts` and move both params into that subsection, but moving `UserAccounts` counts into member part makes probably more sense. After all, there will be also the capacity information stored. Ok, let me change it :-)

MatousJobanek: Check this one: e0448a7 is it better?
I created `CapacityUsage` subsection in both host and member status parts and moved the counts there. Later, I can easily add there the resource usage of the member clusters or also modify the host usage when needed.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -151,16 +166,29 @@ type RegistrationServiceHealth struct {
...
 // Member contains the status of a member cluster
+// +k8s:openapi-gen=true
 type Member struct {
-
 	// The cluster identifier
 	ClusterName string `json:"clusterName"`
 
+	// Capacity usage of the member cluster
+	// +optional
+	CapacityUsage CapacityUsageMember `json:"capacityUsage,omitempty"`
```
alexeykazakov, line 176: It should be a pointer to make it truly optional, right?

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -91,6 +92,10 @@ type HostOperatorStatus struct {
 	// The status of the host operator's deployment
 	DeploymentName string `json:"deploymentName"`
 
+	// Capacity usage of the host cluster
+	// +optional
+	CapacityUsage CapacityUsageHost `json:"capacityUsage,omitempty"`
```
alexeykazakov, line 97: Same here. Should be a pointer.

alexeykazakov:
> I cannot put it into the MemberStatusStatus part because it is part of the MemberStatus

I actually meant to move it to Member. Not MemberStatus. So, what you did in https://github.com/codeready-toolchain/api/pull/182/commits/e0448a7332b732cddb29cbfcc6b9bb44321ec7ae was exactly what I was referring too. Looks good!

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
MatousJobanek: I don't think so. 
I can test it to be 100% sure, but I don't really see any reason for using a pointer for this. The struct contains only the basic type (integer) and it has the `optional`/`omitempty` flags - this means that any existing CR that doesn't contain this parameter should be valid and produce a struct with default values...

alexeykazakov: Let's test it. I know that optional fields with a basic type don't have to be pointers but `CapacityUsageMember` itself is a struct. So, let's test it it make sure we won't break crt-stage if deploy it without using pointers.

MatousJobanek: So I just tested it and the upgrade works like a charm locally. To be honest, I'm not sure why we insist on using pointers for such primitive structs. The `optional`/`omitempty` flags should ensure that the field is optional. Of course, there are some cases when we needed to use the pointer -  for example, previously we had to use it in case of `ClusterResources` that contained a `Template` which is a very complex struct - but that's not the case for `ToolchainStatus` parts.
If I'm not mistaken, the failure of the rolling update happened only when we introduced a "required" parameter that wasn't present in the already existing CRs, but not for optional parameters - otherwise, we would face such an error much often.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -101,7 +106,16 @@ type HostOperatorStatus struct {
 	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 }
 
+// CapacityUsageHost contains information about the capacity usage in host cluster
+// +k8s:openapi-gen=true
+type CapacityUsageHost struct {
```
xcoulon, line 111: Sorry for the late reply, but I find the `CapacityUsage` name to be a bit odd. I believe that "Capacity" refers to the maximum number of resources that the cluster can contain. Should we rename it to `ResourceUsage` instead?  
Just my thoughts...

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
xcoulon: Maybe we had such an issue with the `Template` struct because it contains maps?

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
MatousJobanek: Ah, the naming :exploding_head: 
As I understand the name `CapacityUsage` it shows the current usage of the whole cluster capacity:
* how many users we provisioned
* how much memory is used
* how many PVCs were assigned (or anything else we want to monitor)

The limit of the capacity is then set somewhere else - in configuration file.
The reason why I chose the `Capacity` name and not `Resource` is that the number of provisioned users doesn't have anything in common with resources (memory, cpu, storage). But I can be wrong :man_shrugging:

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
alexeykazakov: Don't remember with what fields/CRD we had issues before but we did have such issues with "optional" structs multiple times before. So it was worth double checking. We will need to introduce an e2e tests for upgrades at some point..

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
xcoulon: > The reason why I chose the `Capacity` name and not `Resource` is that the number of provisioned users doesn't have anything in common with resources (memory, cpu, storage). But I can be wrong 🤷‍♂️

ah right, I missed the CPU, memory, PVC metrics, so maybe `ResourceUsage` is not the best choice either. Although, CPU and memory are resources as well, but it may be confusing with "resources" term which we use a lot to refer to custom resources stored in etcd...

PR #183, opened 2020-09-02
adding timestamp for init verification endpoint and phone-hash label

tinakurian:
## Description
i think there may be an edge case that disallows us to re-use the `verification-timestamp` for both the init verification endpoint as well as the verify code endpoint.
 
as a user, if a user max's out generating new verification codes and the daily limit is reached i would need to wait until the `verification-timestamp` has elapsed 24hrs. then if the user then goes ahead and attempts verifying incorrect codes prior to the `verification-timestamp` elapsing 24hrs,  they will max out the attempts allowed. now, if the user waits until 24hrs has elapsed and invokes the init verification endpoint, we would reset the verification-timestamp and the counter (similar to what is done here): https://github.com/codeready-toolchain/registration-service/blob/master/pkg/verification/verification_service.go#L106-L111

however, when the user attempts to call the verify code endpoint at this point, the verification-timestamp has been reset and hence the max attempts for the verify code endpoint would not be reset.. 

this PR is to add a new timestamp label for the init verification endpoint and to also add the phone-hash label for CRT-800..

please do suggest alternative names for the labels, i don't feel too content with the name i have for the new timestamp

## Checks
yes
no


pkg/apis/toolchain/v1alpha1/usersignup_types.go
```diff
@@ -14,10 +14,14 @@ const (
...
 	// UserSignupUserEmailAnnotationKey is used for the usersignup email annotations key
 	UserSignupUserEmailAnnotationKey = LabelKeyPrefix + "user-email"
+	// UserSignupPhoneNumberLabelKey is used for the usersignup phone number label key
+	UserSignupPhoneNumberLabelKey = LabelKeyPrefix + "phone-number"
 	// UserSignupVerificationCodeAnnotationKey is used for the usersignup verification code annotation key
 	UserSignupVerificationCodeAnnotationKey = LabelKeyPrefix + "verification-code"
 	// UserSignupVerificationTimestampAnnotationKey is used for the usersignup verification timestamp annotation key
 	UserSignupVerificationTimestampAnnotationKey = LabelKeyPrefix + "verification-timestamp"
+	// UserSignupVerificationCodeGeneratedTimestampAnnotationKey is used for the usersignup verification code generated timestamp annotation key
+	UserSignupVerificationCodeGeneratedTimestampAnnotationKey = LabelKeyPrefix + "verification-code-generated-timestamp"
```
sbryzak, line 24: perhaps UserSignupVerificationInitTimestampAnnotationKey

sbryzak (changes requested):
Looks good, but we will need a counter also.  Maybe something like `UserSignupVerificationSentCounterAnnotationKey`

pkg/apis/toolchain/v1alpha1/usersignup_types.go
tinakurian: > Looks good, but we will need a counter also. Maybe something like `UserSignupVerificationSentCounterAnnotationKey`

thanks @sbryzak we already have a counter: https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/usersignup_types.go#L22

tinakurian: updated the name here: d3973f4f75604d4552cb391313ddc1dc89836813

tinakurian:
> Looks good, but we will need a counter also. Maybe something like `UserSignupVerificationSentCounterAnnotationKey`

thanks @sbryzak we already have a counter: https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/usersignup_types.go#L22

sbryzak (approved):

PR #194, opened 2020-10-13
Add Console and Che URLs to member and toolchain status resources

alexeykazakov:
## Description
We need to move Member cluster Web Console and Che Dashboard URLs from MUR resource to Member/Toolchain status resources. This PR is adding the new corresponding fields to the member (directly) and toolchain statuses (indirectly) resources. We will clean up the MUR API later when we move the code to the Status resource controller and update the registration service.

## Checks

yes

yes

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/300
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/207


pkg/apis/toolchain/v1alpha1/memberstatus_types.go
```diff
@@ -45,6 +45,31 @@ type MemberStatusStatus struct {
 	// Resource usage of the cluster
 	// +optional
 	ResourceUsage ResourceUsage `json:"resourceUsage,omitempty"`
+
+	// Routes/URLs of the cluster, such as Console and Che Dashboard URLs
+	// +optional
+	Routes *Routes `json:"routes,omitempty"`
```
MatousJobanek, line 51: I guess that the `Routes` field doesn't have to be a pointer - the default values (in case the routes are not identified in the existing CRs) should be quite obvious - an empty string `""` for URLs and empty array for conditions. 
But I don't insist on it.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -27,9 +27,12 @@ const (
...
 
 	// member status reasons
-	ToolchainStatusMemberStatusNoClustersFoundReason   = "NoMemberClustersFound"
-	ToolchainStatusMemberStatusNotFoundReason          = "MemberStatusNotFound"
-	ToolchainStatusMemberToolchainClusterRemovedReason = "MemberToolchainClusterRemoved"
+	ToolchainStatusMemberStatusNoClustersFoundReason         = "NoMemberClustersFound"
+	ToolchainStatusMemberStatusNotFoundReason                = "MemberStatusNotFound"
+	ToolchainStatusMemberToolchainClusterRemovedReason       = "MemberToolchainClusterRemoved"
+	ToolchainStatusMemberStatusConsoleRouteUnavailableReason = "ConsoleRouteUnavailable"
+	ToolchainStatusMemberStatusCheRouteUnavailableReason     = "CheRouteUnavailable"
```
MatousJobanek, line 34: By setting this reason we define Che as a hard dependency or necessary condition for Dev Sandbox platform. I mean, if I wanted to deploy the operators to some custom clusters where Che is not available, then I won't have a way to make it running - The resulting `MemberStatus` and thus also `ToolchainStatus` won't have `AllComponentsReady` status.

MatousJobanek (commented):
Good idea with putting all routes/URLs together inside the additional type `Routes`, I'm just not sure about using the conditions and relying on Che.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
MatousJobanek: From my point of view, if the Che is not installed (and thus URL is not available), then the `CheDashboardURL` param should be empty and - as a consequnce - the host operator should not set anything it in the MUR status.
I guess that the current implementation does something similar.

pkg/apis/toolchain/v1alpha1/memberstatus_types.go
alexeykazakov: I used a pointer just to be on the safe side :)
We did have a problem with optional fields and not pointers before but I don't know what exactly was different fro other optional fields with no pointers but which worked fine.
That would require some testing.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
alexeykazakov: I'm not sure about Che dependency either.
It would be great to have something for Che routes monitoring but I agree that it should be rather optional.
I would still keep it here as a reason. It's just a const. How it's used is another story. And it's up to the controller to fail the entire status with that reason or use it differently.

MatousJobanek: I understand your point - just don't forget to take care of it in the other PR here: https://github.com/codeready-toolchain/member-operator/pull/208/files#diff-46039918795b512c2c3c08d32a609f0143cea872da1a2caef3592b8334c5c23eR377-R380 and don't set the error when the route is not found.
Would you mind renaming the constant/reason then? For example
```suggestion
	ToolchainStatusMemberStatusUnableToGetCheRouteReason     = "UnableToGetCheRoute"
```
This message more shows that the problem is with getting, not with the existence.

pkg/apis/toolchain/v1alpha1/memberstatus_types.go
MatousJobanek: Sure, I know why you did it - that's why I said that I didn't insist on it.
If I'm not mistaken, the problem was mainly with the required values, for optional it was with the types where the default value is not obvious - for example `Time`, or very complex ones like `Template`

MatousJobanek (approved):
Thanks for answering my concerns/questions.
Please consider renaming the reason constant.

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
alexeykazakov: Well.. I was thinking of using this reason when/if we want to report any problem with accessing the che dashboard route. Without specifics.
Whether we want to set the entire ready status condition to false with this reason if there is a problem accessing the Che route in the controller or not is a separate question.
There are pros and cons. Maybe we should keep it optional.. Let's say by default our configuration doesn't require Che to be available but we can make it required in our configuration so the controller starts using the reason and fail the entire ready condition if so.

alexeykazakov: @MatousJobanek I'm merging this PR but let's keep discussing it. This constant does not affect CRDs anyway so it's pretty easy to change it anytime.

PR #196, opened 2020-10-14
chore: drop zones and region from ToolchainCluster

MatousJobanek:
## Description
Drop zones and region from ToolchainCluster CR because:
* it doesn't bring any additional value
* requires additional cluster-scoped permission to access `Nodes`
* makes the stuff a bit simpler
* if we are interested in Node zones & region, we should put it into Member/ToolchainStatus CRs

## Checks

yes

yes

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/303
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/210


tinakurian (approved):

alexeykazakov (approved):
+1

PR #220, opened 2021-03-12
Added new constants for user pre-deactivation notification

sbryzak:
## Description
This PR adds constants for a new "deactivating" UserSignup condition in order to manage the notifications sent to users prior to them becoming deactivated.  Only constants have been added, no structural changes to CRDs.

Fixes https://issues.redhat.com/browse/CRT-994

## Checks

2. Does `make generate` change anything in other projects (host-operator, member-operator, toolchain-common)? **[no]** 

### DIFF
--- a/pkg/apis/toolchain/v1alpha1/usersignup_types.go
+++ b/pkg/apis/toolchain/v1alpha1/usersignup_types.go
@@ -11,6 +11,9 @@ const (
 	UserSignupApproved ConditionType = "Approved"
 	// UserSignupComplete means provisioning is complete
 	UserSignupComplete ConditionType = "Complete"
+	// UserSignupUserDeactivatingNotificationCreated is used to track the status of the notification send to a user
+	// shortly before their account is due for deactivation
+	UserSignupUserDeactivatingNotificationCreated ConditionType = "UserDeactivatingNotificationCreated"
 	// UserSignupUserDeactivatedNotificationCreated means that the Notification CR was created so the user should be notified about their deactivated account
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):

pkg/apis/toolchain/v1alpha1/usersignup_types.go
```diff
@@ -11,6 +11,9 @@ const (
 	UserSignupApproved ConditionType = "Approved"
 	// UserSignupComplete means provisioning is complete
 	UserSignupComplete ConditionType = "Complete"
+	// UserSignupUserDeactivatingNotificationCreated is used to track the status of the notification send to a user
+	// shortly before their account is due for deactivation
+	UserSignupUserDeactivatingNotificationCreated ConditionType = "UserDeactivatingNotificationCreated"
```
MatousJobanek, line 16: Wouldn't be either `UserBeforeDeactivationNotificationCreated` or  `UserPreDeactivationNotificationCreated` more accurate? But I'm fine with your version as well :+1:

MatousJobanek (approved):
Thanks for separating the reason constants and adding comments to them - This looks much better :+1:

pkg/apis/toolchain/v1alpha1/usersignup_types.go
sbryzak: I originally went with "Pre" however I didn't like the look of it - our constant names are already huge as it is and "deactivating" made sense in the context of notifying the user "hey, your account will be deactivating shortly".

PR #224, opened 2021-04-06
metrics: add field to store metrics in ToolchainStatus

xcoulon:
also, add following constants:
- `UsersPerActivationMetricKey` to store the current "Users per Activations"
  counter in the ToolchainStatus metrics
- `UserSignupActivationCounterAnnotationKey` to track the number
  of activations of a user, using an annotation on her `UserSignup` resource.

Fixes https://issues.redhat.com/browse/CRT-988


pkg/apis/toolchain/v1alpha1/usersignup_types.go
```diff
@@ -75,6 +79,8 @@ const (
 	UserSignupMissingEmailHashLabelReason          = "MissingEmailHashLabel"
 	UserSignupInvalidEmailHashLabelReason          = "InvalidEmailHashLabel"
 	UserSignupVerificationRequiredReason           = "VerificationRequired"
+	UserSignupUnableToAddFinalizerReason           = "UnableToAddFinalizer"
+	UserSignupUnableToRemoveFinalizerReason        = "UnableToRemoveFinalizer"
```
alexeykazakov, line 83: Do you need this too?

alexeykazakov (approved):

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -82,6 +85,8 @@ type ToolchainStatusStatus struct {
 	// +listMapKey=clusterName
 	Members []Member `json:"members,omitempty" patchStrategy:"merge" patchMergeKey:"clusterName"`
 
+	Metrics map[string]string `json:"metrics,omitempty"`
```
alexeykazakov, line 88: This should be optional, right? Also, shouldn't we specify the map type / merge strategy?

alexeykazakov (changes requested):
I have a couple of questions to clarify...

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
xcoulon: good point. Let me add that

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -82,6 +85,12 @@ type ToolchainStatusStatus struct {
 	// +listMapKey=clusterName
 	Members []Member `json:"members,omitempty" patchStrategy:"merge" patchMergeKey:"clusterName"`
 
+	// Metrics is a map that stores (seralized) metrics to be exposed on Prometheus.
+	// +optional
+	// +mapType=atomic
+	// +patchStrategy=merge
+	Metrics map[string]string `json:"metrics,omitempty" patchStrategy:"merge"`
```
MatousJobanek, line 92: I'm sorry for my dumb question - what is the expected content of this map?
From the code I would assume:
```
status:
  metrics:
    UsersPerActivation: '{"1":2,"2":2,"3":1,"4":1,"5":1,"6":1,"7":1,"8":1}'
```
if so,
* shouldn't be the first letter of `UsersPerActivation` lower-case - as the rest of the keys? `usersPerActivation`
* why storing the values in the json format and dealing with marshaling & unmarshaling? Is there any benefit?

MatousJobanek: I mean, why not storing it as a map?

xcoulon: we can't put `interface{}` values, so I opted for a string by serializing the JSON content

xcoulon: > I'm sorry for my dumb question - what is the expected content of this map?
> From the code I would assume:
> 
> ```
> status:
>   metrics:
>     UsersPerActivation: '{"1":2,"2":2,"3":1,"4":1,"5":1,"6":1,"7":1,"8":1}'
> ```
> 
> if so,
> 
> * shouldn't be the first letter of `UsersPerActivation` lower-case - as the rest of the keys? `usersPerActivation`
ok, I can do that

xcoulon: > > I'm sorry for my dumb question - what is the expected content of this map?
> > From the code I would assume:
> > ```
> > status:
> >   metrics:
> >     UsersPerActivation: '{"1":2,"2":2,"3":1,"4":1,"5":1,"6":1,"7":1,"8":1}'
> > ```
> > 
> > 
> > if so,
> > 
> > * shouldn't be the first letter of `UsersPerActivation` lower-case - as the rest of the keys? `usersPerActivation`
> 
> ok, I can do that

done in bf67281

MatousJobanek: I meant something like:
```
map[string]map[string]int
```

pkg/apis/toolchain/v1alpha1/usersignup_types.go
xcoulon: removed since we decided not to decrease the counter and metrics when a user is deleted

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
xcoulon: done in 9f6680e

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
xcoulon: > I meant something like:
> 
> ```
> map[string]map[string]int
> ```

the code generation fails with the following error: 
```
github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go:92:21: map values must be a named type, not *ast.MapType
```

xcoulon: actually, the code generation does not fail if I use something like 
```
map[string]Metric
```
with 
```
type Metric map[string]int 
```
so let me see if this solution can work up to the E2E tests

xcoulon: done in 31fcd86 (and subsequently in https://github.com/codeready-toolchain/host-operator/pull/401/commits/04b0eca1e7343f4ad7fd943c747ad2e33625df7c) and it worked with the E2E tests on my temp cluster

alexeykazakov (approved):

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
```diff
@@ -82,6 +85,12 @@ type ToolchainStatusStatus struct {
 	// +listMapKey=clusterName
 	Members []Member `json:"members,omitempty" patchStrategy:"merge" patchMergeKey:"clusterName"`
 
+	// Metrics is a map that stores (seralized) metrics to be exposed on Prometheus.
```
MatousJobanek, line 88:
```suggestion
	// Metrics is a map that stores metrics to be exposed on Prometheus.
```

MatousJobanek (approved):
Awesome! Thanks a lot for changing the type of the stored value :+1:

pkg/apis/toolchain/v1alpha1/toolchainstatus_types.go
xcoulon: indeed, this comment needs to be updated now ;)

PR #234, opened 2021-05-17
fix: rename UserID->Userid & add missing listType

MatousJobanek:
## Description
Renames `UserID` to Userid`
adds missing listType to states

## Checks

**yes**

**yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/436

### DIFF
--- a/pkg/apis/toolchain/v1alpha1/usersignup_types.go
+++ b/pkg/apis/toolchain/v1alpha1/usersignup_types.go
@@ -143,7 +143,7 @@ type UserSignupSpec struct {
 	Approved bool `json:"approved,omitempty"`
 
 	// The user's user ID, obtained from the identity provider from the 'sub' (subject) claim
-	UserID string `json:"userid"`
+	Userid string `json:"userid"`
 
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):

pkg/apis/toolchain/v1alpha1/usersignup_types.go
```diff
@@ -143,7 +143,7 @@ type UserSignupSpec struct {
 	Approved bool `json:"approved,omitempty"`
 
 	// The user's user ID, obtained from the identity provider from the 'sub' (subject) claim
-	UserID string `json:"userid"`
+	Userid string `json:"userid"`
```
xcoulon, line 146: I guess we could restore the `UserID` type field, mapped to a `userID` JSON field, but that would probably require multiple migration steps 😬

MatousJobanek: Yes, exactly - we would need to migrate all current CRs

PR #241, opened 2021-06-01
chore: drop vendoring and get tools with go

MatousJobanek:
## Description
I decided to drop vendoring in order to get the needed version of the "generate" tools and start using a smarter version of what is generated by operator-sdk. It gets the binaries via `go get` command and stores them in `/bin` directory together with a file that contains the used versions of the binaries - to make sure that the correct version is used.

The main reason for doing it is that we will need the `kubebuilder` binary also in host/member operator repositories and keeping the vendor folder is sometimes pretty annoying. So, the same approach will be used in host/member operator repos to get `kubebuilder` & `kustomize` binaries.

I know that we agreed on not installing binaries from repos in makefiles targets, but - to be honest - I don't see any difference between getting the binaries via vendor folder and getting them via `go get` command.

Anyway, I'm open to discussion. 

Surprisingly, there are also some changes in the generated CRDs:
https://github.com/codeready-toolchain/member-operator/pull/260
https://github.com/codeready-toolchain/host-operator/pull/452

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ build/_output
 # the 'config' (originally deploy) directory should not be committed: all generated CRD files should be dispatched in the 'host-operator' and 'member-operator' repositories
 config
 deploy
+
+/bin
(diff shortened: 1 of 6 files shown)


alexeykazakov (approved):
I think this approach is fine.

rajivnathan (approved):
👍

PR #243, opened 2021-06-14
Toolchainconfigstatus

rajivnathan:
## Description
Adds a status condition and sync errors map to the ToolchainConfig CR.

## Checks
1. Have you run `make generate` target? **yes**

2. Does `make generate` change anything in other projects (host-operator, member-operator, toolchain-common)? **yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/460


alexeykazakov (approved):

MatousJobanek (approved):

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -97,6 +108,21 @@ type Deactivation struct {
 // ToolchainConfigStatus defines the observed state of ToolchainConfig
 // +k8s:openapi-gen=true
 type ToolchainConfigStatus struct {
+
+	// SyncErrors is a map of sync errors indexed by toolchaincluster name that indicates whether
+	// an attempt to sync configuration to a member cluster failed
+	// +optional
+	// +mapType=atomic
+	SyncErrors map[string]string `json:"syncErrors,omitempty"`
```
MatousJobanek, line 116: @rajivnathan Just out of curiosity - why adding an extra field for the errors and not putting then into the message of the condition?

rajivnathan: I just thought it could be too congested if there are multiple sync errors. eg. if we have 3 member clusters and syncing failed to all then the message would be long and perhaps a bit hard to read right away.

MatousJobanek: ok, I'm fine with this structure - we can always move it to condition message if needed :+1:

alexeykazakov: Don't forget to cleanup the errors. Just updating the condition wouldn't remove the errors from the status :)

PR #245, opened 2021-06-15
feat: migrate api to operator-sdk 1.8.0

MatousJobanek:
## Migrates api repo to operator-sdk 1.8.0
* updates dependencies to use k8s v0.20.2
* upgrades go to 1.16
* modifies generate targets so
    * it doesn't regenerate CSV
    * dispatches CRDs to the new location
    * uses latest openapi-gen binary
* rewrites scripts so they can work with the new structure and latest operator-sdk commands - some of the scripts were removed as they are not needed anymore
* adds `install-operator.sh` which will install the needed operator - is also used in e2e tests

### TODO in the next PR:
* move scripts to toolchain-cicd repo


alexeykazakov (approved):

xcoulon:
> * adds `install-operator.sh` which will install the needed operator - is also used in e2e tests

why do we need this script in this repo, and not just in https://github.com/codeready-toolchain/toolchain-e2e ?

MatousJobanek:
>why do we need this script in this repo, and not just in https://github.com/codeready-toolchain/toolchain-e2e ?

because it's called from the host/operator repos - from the target `install-current-operator`. But it doesn't matter where it is because - as I already mentioned it - I want to move all scripts to toolchain-cicd repo

xcoulon:
> > why do we need this script in this repo, and not just in https://github.com/codeready-toolchain/toolchain-e2e ?
> 
> because it's called from the host/operator repos - from the target `install-current-operator`. But it doesn't matter where it is because - as I already mentioned it - I want to move all scripts to toolchain-cicd repo

ok, thanks for the clarification. I wasn't sure if this script was supposed to be moved to the toolchain-cicd repo.

PR #247, opened 2021-06-29
Modified Notification CRD to add Context, deprecate UserID

sbryzak:
## Description
This PR deprecates the `UserID` property of the `Notification` CRD spec, and replaces it with a new `Context` which will allow more flexible provision of custom values for the generation of notification content.

API changes for https://issues.redhat.com/browse/CRT-821

## Checks

2. Does `make generate` change anything in other projects (host-operator, member-operator, toolchain-common)? **[yes]** 

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/476


MatousJobanek (commented):
This looks good, but shouldn't you also deprecate `Subject` and `Content`?
IIRC, the plan is to have only these parameters inside NotificationSpec:
```
Recipient string 
Template string
Context map[string]string 
```

api/v1alpha1/notification_types.go
```diff
@@ -42,10 +42,17 @@ type NotificationSpec struct {
...
-	// Recipient may be used as an alternative to UserID to specify an email address where the notification will be delivered.
-	Recipient string `json:"recipient,omitempty"`
+	// Recipient is used to specify the email address where the notification will be delivered.  It must comply with
+	// section 3.4.1 of RFC2822, and should be formatted to include the user's first and last names,
+	// e.g. "John Smith <jsmith@example.com>"
+	Recipient string `json:"recipient"`
+
+	// Context is used to set a number of arbitrary values to be passed to the notification content text formatter,
+	// for inclusion in the body of the notification.
+	Context map[string]string `json:"context"`
```
alexeykazakov, line 55: Shouldn't we add `//+optional` for both `Context` and `UserID`?

sbryzak: Definitely for `UserID`, yes... for `Context` i guess you could have a notification template that doesn't require any data binding.  I've updated both.

sbryzak:
@MatousJobanek we still need `Subject` and `Content` in order to generate ad-hoc notifications (i.e. those that don't have a predefined template).  The new `Context` map will just hold the values to be merged into the actual notification content.

api/v1alpha1/notification_types.go
alexeykazakov: But even regardless of what we could have in the future, right now we our notification resources do not have Context. It's a new API. So we have to make it optional so we do not break the operator update by introducing required API which is missing in the existing resources.

alexeykazakov (approved):

MatousJobanek:
sorry for the late response @sbryzak 

> we still need Subject and Content in order to generate ad-hoc notifications (i.e. those that don't have a predefined template).

in the design doc, we agreed on dropping `Subject` and `Content`. There are two reasons for that:
1. When we need to introduce a new "ad-hoc" notifications, we still need to do it in the code - that means we need to make also a new release/deployment. There is no reason for doing it only in the code and not via the templates.
2. The main reason for having `Subject` and `Content` was the lack of the generic context that you just introduced. We needed that for the emails for admins - in that case we don't have any UserId that was/is required by the template processing.

PR #253, opened 2021-08-11
Add replicas to reg service config

rajivnathan:
## Description
Adds replicas to registration service config and organizes the fields in alphabetical order.

## Checks
1. Have you run `make generate` target? **[yes/no]** yes

2. Does `make generate` change anything in other projects (host-operator, member-operator, toolchain-common)? **[yes/no]**  yes

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/495


api/v1alpha1/toolchainconfig_types.go
```diff
@@ -231,20 +239,17 @@ type RegistrationServiceConfig struct {
...
 
-	// Keeps parameters necessary for the registration service analytics config
+	// Replicas specifies the number of replicas to use for the registration service deployment
 	// +optional
-	Analytics RegistrationServiceAnalyticsConfig `json:"analytics,omitempty"`
-
-	// Keeps parameters necessary for the registration service authentication config
-	// +optional
-	Auth RegistrationServiceAuthConfig `json:"auth,omitempty"`
+	Replicas *string `json:"replicas,omitempty"`
```
alexeykazakov, line 252: Shouldn't it be an int32 with `1` as the default value?
```suggestion
	Replicas *int32 `json:"replicas,omitempty"`
```

rajivnathan: yes, for some reason I thought we used string but int32 makes more sense

rajivnathan: Regenerated the files and updated https://github.com/codeready-toolchain/host-operator/pull/495 as well

alexeykazakov (approved):
Looks good. But I thought it would be useful to mention the default values in API?

rajivnathan:
@alexeykazakov we set the default in the getters for ToolchainConfig and MemberOperatorConfig. eg. https://github.com/codeready-toolchain/member-operator/blob/34b134adf9572fae2d4769867be0d48c40b673ea/controllers/memberoperatorconfig/configuration.go#L164 so maybe it would be better to just add comments to those functions for defaults?

alexeykazakov:
@rajivnathan yeah I know. I was thinking that it could be helpful to mention the default values in the API. In the comments I mean.
But since clients (our controllers) do not actually use the API directly and always use the `Configuration` struct for that then I think that's fine if we don't mention the default values in the API.

PR #255, opened 2021-08-26
Add ToolchainConfig status condition type and reasons for reg service

rajivnathan:
## Description
The ToolchainConfig controller will be deploying the registration service. This PR adds the ConditionType and Reasons to use for ToolchainConfig conditions related to deploying the registration service.

Related PR: https://github.com/codeready-toolchain/host-operator/pull/507

## Checks
yes

No


rajivnathan:
@xcoulon fair question. It's because the ToolchainStatus will cause a notification to be sent to us to inform of a problem but not necessarily what caused it. For example, in the ToolchainConfig status the registration service deploy condition can be failed and have a message like `failed to process registration service template` and include the cause error as well. So it should help to debug problems.

MatousJobanek (approved):

PR #263, opened 2021-11-10
refactor: UserAccount.Spec.NSTemplateSet as optional

xcoulon:
See https://github.com/codeready-toolchain/host-operator/pull/537 and https://github.com/codeready-toolchain/member-operator/pull/309


alexeykazakov (approved):
Looks good but do you have a PR for the member cluster repo?

MatousJobanek:
There should be PR not only for member-operator but also for host-operator. This change affects both operators.

alexeykazakov:
> There should be PR not only for member-operator but also for host-operator. This change affects both operators.

How does it affect the host operator? This CRD is installed by the member operator, right?

MatousJobanek:
yes, but the `NSTemplateSet` parameter that is modified is defined in `UserAccountSpecBase` type which is used in `MUR.spec.[]UserAccountEmbedded.UserAccountSpecEmbedded`. And since the parameter is changed to a pointer, then I would expect that the code would need some additional changes - UserSignup, ChangeTierRequest & Deactivation controller use that.

xcoulon:
> yes, but the `NSTemplateSet` parameter that is modified is defined in `UserAccountSpecBase` type which is used in `MUR.spec.[]UserAccountEmbedded.UserAccountSpecEmbedded`. And since the parameter is changed to a pointer, then I would expect that the code would need some additional changes - UserSignup, ChangeTierRequest & Deactivation controller use that.

ah right, I forgot about that! 🤦‍♂️ 
Let me prepare a PR for host-operator as well

MatousJobanek:
btw, wouldn't be just easier to only mark it as optional and check if the TierName is set on the member side? As I proposed in the Jira story?

xcoulon:
> btw, wouldn't be just easier to only mark it as optional and check if the TierName is set on the member side? As I proposed in the Jira story? You know - don't change it to pointer

To be honest, I thought that marking as optional also implied using a pointer. 
Now, since I already did the work, can we consider this solution?

MatousJobanek:
> To be honest, I thought that marking as optional also implied using a pointer.

nope, you can easily make a field optional even if it's not a pointer. That applies also to the structs. There are sometimes complications with huge complicated types like `Template` or with a special type like `Time`.

>  Now, since I already did the work, can we consider this solution?

yes, sure - feel free to go ahead with your solution if you have already implemented all the changes. I just thought that it could make the changes smaller/easier. But it's fully up to you :wink:

alexeykazakov:
Ah... I didn't notice that Xavier changed it to a pointer...
Thought he only made it optional. So yeah in this case code changes might be required.

xcoulon:
> Ah... I didn't notice that Xavier changed it to a pointer... Thought he only made it optional. So yeah in this case code changes might be required.

yes, I've addressed them today in https://github.com/codeready-toolchain/host-operator/pull/537 as well as https://github.com/codeready-toolchain/toolchain-e2e/pull/407 (plus updates/fixes in https://github.com/codeready-toolchain/toolchain-common/pull/216)

MatousJobanek (approved):

PR #264, opened 2021-11-18
Added constants for ToolchainEvent state label

sbryzak:
## Description
This PR adds constants used to manage the ToolchainEvent's state label.  No regeneration necessary, as PR only contains new constant definitions and nothing else.

## Checks
1. Have you run `make generate` target? **[no]**

2. Does `make generate` change anything in other projects (host-operator, member-operator)? **[no]**


sbryzak:
Will approach this in a different way.

PR #265, opened 2021-11-18
Removed ActivationCode from Spec, added constant for activation-code label

sbryzak:
## Description
This PR removes the `ActivationCode` property from `ToolchainEvent.Spec`, and instead introduces a new label constant so that the activation code can instead be stored as a label value.  This will enable easier lookup of the `ToolchainEvent` resource.

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/538

### DIFF
--- a/api/v1alpha1/toolchainevent_types.go
+++ b/api/v1alpha1/toolchainevent_types.go
@@ -8,6 +8,8 @@ const (
 
 	// Status condition reasons
 	ToolchainEventInvalidTierReason = "InvalidTier"
+
+	ToolchainEventActivationCodeLabelKey = LabelKeyPrefix + "activation-code"
 )
 
 // ToolchainEventSpec defines the parameters for a Toolchain event, such as a training session or workshop. Users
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):

MatousJobanek:
> This will enable easier lookup of the ToolchainEvent resource.

Just out of curiosity - do we really expect to have so many ToolchainEvent resources so it would be too expensive to list them all and go over their Spec?
The Spec is easier to read & maintain and I think that there will be only a few of these resources in the cluster. 
I'm just asking and sharing my opinion - functional-wise, the label will work for sure, so I'm not blocking it. I just wanted to be clear that this is what we want.

sbryzak:
@MatousJobanek I had this very same discussion with Alexey, and we came to the agreement that it would be more convenient to store the activation code in a label.  One thing that we want to be able to do is allow the admin user who creates the event in the first place to not have to enter an activation code, in which case the controller will generate one automatically, and we agreed that having a label for this might be a better option.

MatousJobanek (approved):
Right, that generation function is a really good idea :+1: 
In that case, I agree that using a label is a much better option for that

PR #278, opened 2021-12-13
feat: reason for Space status update failures

xcoulon:
Used in https://github.com/codeready-toolchain/host-operator/pull/555

### DIFF
--- a/api/v1alpha1/changetierrequest_types.go
+++ b/api/v1alpha1/changetierrequest_types.go
@@ -18,7 +18,7 @@ const (
 
 	// Status condition reasons
 	ChangeTierRequestChangedReason       = "Changed"
-	ChangeTierRequestChangeFiledReason   = "ChangeFailed"
+	ChangeTierRequestChangeFailedReason  = "ChangeFailed"
 	ChangeTierRequestDeletionErrorReason = "UnableToDeleteChangeTierRequest"
 )
 
--- a/api/v1alpha1/space_types.go
+++ b/api/v1alpha1/space_types.go
@@ -24,6 +24,7 @@ const (
 	SpaceTerminatingFailedReason           = terminatingFailedReason
 	SpaceUpdatingReason                    = updatingReason
 	SpaceNSTemplateSetUpdateFailedReason   = "NSTemplateSetUpdateFailed"
+	SpaceStatusUpdateFailedReason = "UnableToUpsateStatus"
 )
 
 // SpaceSpec defines the desired state of Space
--- a/api/v1alpha1/space_types.go
+++ b/api/v1alpha1/space_types.go
@@ -24,7 +24,7 @@ const (
 	SpaceTerminatingFailedReason           = terminatingFailedReason
 	SpaceUpdatingReason                    = updatingReason
 	SpaceNSTemplateSetUpdateFailedReason   = "NSTemplateSetUpdateFailed"
-	SpaceStatusUpdateFailedReason = "UnableToUpsateStatus"
+	SpaceStatusUpdateFailedReason          = "UnableToUpsateStatus"
 )
 
 // SpaceSpec defines the desired state of Space


api/v1alpha1/space_types.go
```diff
@@ -24,6 +24,7 @@ const (
 	SpaceTerminatingFailedReason           = terminatingFailedReason
 	SpaceUpdatingReason                    = updatingReason
 	SpaceNSTemplateSetUpdateFailedReason   = "NSTemplateSetUpdateFailed"
+	SpaceStatusUpdateFailedReason = "UnableToUpsateStatus"
```
alexeykazakov, line 27: Formatting seems to be off.

api/v1alpha1/changetierrequest_types.go
```diff
@@ -18,7 +18,7 @@ const (
 
 	// Status condition reasons
 	ChangeTierRequestChangedReason       = "Changed"
-	ChangeTierRequestChangeFiledReason   = "ChangeFailed"
+	ChangeTierRequestChangeFailedReason  = "ChangeFailed"
```
alexeykazakov, line 21: I don't see any links to the paired PRs with updated const (host, member, etc...)

xcoulon: the change is used in https://github.com/codeready-toolchain/host-operator/pull/555. I'll add this link in the description

api/v1alpha1/space_types.go
xcoulon: ah right, and we don't have a linter for this repo :/

xcoulon: well, there's a PR for this now: https://github.com/codeready-toolchain/api/pull/279

xcoulon: fixed in 4101142

api/v1alpha1/changetierrequest_types.go
alexeykazakov: If you extract the related change from 555 to a new PR then we would be able to merge both api and the host change :)
Otherwise this PR will be blocked until 555 is fully ready to be merged. Otherwise we risk blocking api.

xcoulon: yes, I know. I'm fine with keeping this PR blocked until https://github.com/codeready-toolchain/host-operator/pull/555 but since this change isn't related to https://github.com/codeready-toolchain/host-operator/pull/555, I will extract it to another pair of PRs.

xcoulon: see https://github.com/codeready-toolchain/api/pull/282

api/v1alpha1/space_types.go
```diff
@@ -24,6 +24,7 @@ const (
 	SpaceTerminatingFailedReason           = terminatingFailedReason
 	SpaceUpdatingReason                    = updatingReason
 	SpaceNSTemplateSetUpdateFailedReason   = "NSTemplateSetUpdateFailed"
+	SpaceStatusUpdateFailedReason          = "UnableToUpsateStatus"
```
MatousJobanek, line 27:
```suggestion
	SpaceStatusUpdateFailedReason          = "UnableToUpdateStatus"
```

MatousJobanek (commented):
There is a small typo, but why do you need a reason at all?
If it fails once while updating the status, then it will probably fail again when you would try to set the reason to `UnableToUpdateStatuts` in the same status. Or am I missing something?

xcoulon:
> There is a small typo, but why do you need a reason at all? If it fails once while updating the status, then it will probably fail again when you would try to set the reason to `UnableToUpdateStatuts` in the same status. Or am I missing something?

ah, that's a good point indeed! 🤦‍♂️

xcoulon:
> There is a small typo, but why do you need a reason at all? If it fails once while updating the status, then it will probably fail again when you would try to set the reason to `UnableToUpdateStatuts` in the same status. Or am I missing something?

let me close this PR, then. We don't need this reason indeed

PR #279, opened 2021-12-14
chore: linter as a GitHub Action

xcoulon:
Run Go and YAML linters on GitHub Actions

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,10 +1,24 @@
 service:
-  project-path: github.com/codeready-toolchain/api
+  project-path: github.com/codeready-toolchain/host-operator
+  prepare: # see https://github.com/golangci/golangci/wiki/Configuration#config-directives
+    - make generate
 
 run:
   # timeout for analysis, e.g. 30s, 5m, default is 1m
   timeout: 10m
 
+
 linters:
   enable:
-    - gofmt
+    - gofmt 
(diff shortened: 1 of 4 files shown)


xcoulon:
also, I wonder if we need the `ci/prow/build` and `ci/prow/test` jobs at all: I don't think that the `go build github.com/codeready-toolchain/api/api/v1alpha1/` command builds anything at all since there is no `main.go` in that directory, and we don't have any test at all either. In anycase, if we really want to keep those jobs, I would suggest that we migrate them to GitHub Actions, along with the linter proposed in this PR. 

WDYT?

MatousJobanek (approved):
I'm fine with dropping build as well as test jobs. Previously, we had some tests there, but they were removed.

xcoulon:
> I'm fine with dropping build as well as test jobs. Previously, we had some tests there, but they were removed.

great, thanks for your feedback! I will take care of removing them

.golangci.yml
```diff
@@ -1,10 +1,24 @@
 service:
-  project-path: github.com/codeready-toolchain/api
+  project-path: github.com/codeready-toolchain/host-operator
```
rajivnathan, line 2: Was this a mistake?

xcoulon: ah damn'it, I missed that :/

PR #292, opened 2022-02-07
feat: SpaceRoles in NSTemplateTier

xcoulon:
See also https://github.com/codeready-toolchain/host-operator/pull/591

Fixes https://issues.redhat.com/browse/CRT-1345


api/v1alpha1/nstemplatetier_types.go
```diff
@@ -28,6 +28,11 @@ type NSTemplateTierSpec struct {
 	// +optional
 	ClusterResources *NSTemplateTierClusterResources `json:"clusterResources,omitempty"`
 
+	// the templates to set the spaces roles, indexed by role
+	// +patchStrategy=merge
+	// +mapType=atomic
+	SpaceRoles map[string]NSTemplateTierSpaceRoles `json:"spaceRoles,omitempty" patchStrategy:"merge"`
```
MatousJobanek, line 34: does it really make sense to specify `patchStrategy` to `merge` when the type of the map is supposed to be `atomic`?

In addition, we should mark it as optional (at least for the beginning)

xcoulon: ah right, it was supposed to be `optional` but after some changes, I removed that annotation by accident 🤦‍♂️ 
as for the `+patchStrategy=merge` vs `+mapType=atomic`, indeed, it's contradictory. 
See 4c9cd46

note that there are other places where the combination of `+patchStrategy=merge` along with `+mapType=atomic` is used:
- https://github.com/codeready-toolchain/api/blob/master/api/v1alpha1/templateupdaterequest_types.go#L60-L62
- https://github.com/codeready-toolchain/api/blob/master/api/v1alpha1/toolchainstatus_types.go#L99-L101

it's probably worth fixing them, but maybe in a separate PR?

api/v1alpha1/nstemplatetier_types.go
```diff
@@ -45,6 +50,12 @@ type NSTemplateTierClusterResources struct {
 	TemplateRef string `json:"templateRef"`
 }
 
+// NSTemplateTierSpaceRoles the space roles definition in an NSTemplateTier resource
+type NSTemplateTierSpaceRoles struct {
```
MatousJobanek, line 54: btw, the name should have been singular, and not plural 
```suggestion
type NSTemplateTierSpaceRole struct {
```

xcoulon: changed in cdea401

MatousJobanek (approved):

PR #302, opened 2022-04-20
Create UserTier

rajivnathan:
## Description
Creates a new UserTier CRD that will contain user-level configuration. eg. How long a user's account can be active before automatic deactivation.

## Checks
1. Did you run `make generate` target? **[yes]**

2. Did `make generate` change anything in other projects (host-operator, member-operator)? **[yes]**

3. In case of **new** CRD, did you also update make/generate.mk in this repository and `resources/setup/roles/host.yaml` in the sandbox-sre repository? **yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/630
    - sandbox-sre: https://github.com/codeready-toolchain/sandbox-sre/pull/596
    - toolchain-cicd: https://github.com/codeready-toolchain/toolchain-cicd/pull/62


api/v1alpha1/usertier_types.go
```diff
@@ -0,0 +1,78 @@
...
+	// +listMapKey=type
+	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// Updates is an array of all UserTier updates
+	// +optional
+	// +patchMergeKey=startTime
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=startTime
+	Updates []UserTierHistory `json:"updates,omitempty" patchStrategy:"merge" patchMergeKey:"startTime"`
```
MatousJobanek, line 56: what this will be used for? I don't think that we need to keep the history of the UserTiers.
I know that we have it in NSTemplateTier, but in that case, we need to update all Spaces & NSTemplateSets - we don't need to do anything like that in the case of UserTiers.

MatousJobanek (commented):
Looks good, but I think that we can keep the `UserTierStatus` empty

rajivnathan:
@MatousJobanek thanks for the review and good point about `UserTierStatus`. I've removed the update history.

api/v1alpha1/usertier_types.go
```diff
@@ -0,0 +1,70 @@
...
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Conditions is an array of current UserTier conditions
+	// Supported condition types: ConditionReady
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
MatousJobanek, line 48: let's drop the conditions too - we won't use them

api/v1alpha1/usertier_types.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// UserTierHistory a track record of an update
+type UserTierHistory struct {
+	// StartTime is the time when the UserTier was updated
+	UpdateTime metav1.Time `json:"startTime"`
+	// Hash the hash matching of the UserTier spec
+	Hash string `json:"hash"`
+}
```
MatousJobanek, line 57: :ghost: this should be removed as well

MatousJobanek (approved):

api/v1alpha1/usertier_types.go
rajivnathan: Just realized that when I was removing the conditions but you already noticed 😄

api/v1alpha1/usertier_types.go
rajivnathan: Dropped https://github.com/codeready-toolchain/api/pull/302/commits/9b426a7e51372a09867da6b199a836bc5ab5d028

PR #320, opened 2022-09-02
Provisioning user signup

ranakan19:
## Description
adding a condition type and condition reason to indicate usersignup provisioning is in progress.

## Checks

4. In case other projects are changed, please provides PR links. - **NA**

### DIFF
--- a/api/v1alpha1/usersignup_types.go
+++ b/api/v1alpha1/usersignup_types.go
@@ -79,6 +79,7 @@ const (
 	UserSignupUnableToDeleteMURReason              = "UnableToDeleteMUR"
 	UserSignupUnableToCreateSpaceReason            = "UnableToCreateSpace"
 	UserSignupUnableToCreateSpaceBindingReason     = "UnableToCreateSpaceBinding"
+	UserSignupProvisioningSpace                    = "ProvisioningSpace"
 
 	// The UserSignupUserDeactivatingReason constant will be replaced with UserSignupDeactivationInProgressReason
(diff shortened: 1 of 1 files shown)


alexeykazakov (approved):

api/v1alpha1/usersignup_types.go
```diff
@@ -11,6 +11,8 @@ const (
 	UserSignupApproved ConditionType = "Approved"
 	// UserSignupComplete means provisioning is complete
 	UserSignupComplete ConditionType = "Complete"
+	//UserSignupProvisioning means provisioning is in progress
+	UserSignupProvisioning ConditionType = "Provisioning"
```
MatousJobanek, line 15: we don't need a new condition type for this particular reason - it should have been set in the `Complete` one

MatousJobanek (changes requested):
Please drop the new ConditionType - we don't need that

api/v1alpha1/usersignup_types.go
ranakan19: yeah that makes sense! I'll just keep the reason

alexeykazakov: Ah, right! I overlooked that. I was thinking you Kanika are introducing a new reason. Not condition!

ranakan19: done in ec82f43

MatousJobanek (approved):
Thanks

PR #332, opened 2022-12-20
feat: add 'Subject' field in UserSignup

xcoulon:
This field will be populated with the 'sub' claim of the user's access token,
and the existing 'UserID' field will be filled with the 'userID' claim.

As a result, custom resource fields will be aligned with access token claims.


alexeykazakov:
This is outdated. Shane is working on addressing it.

PR #333, opened 2022-12-22
add SpaceRequest CRD

mfrancisc:
## Description
Add new SpaceRequest CRD

See Jira ticket: https://issues.redhat.com/browse/ASC-244

## Checks

    - `resources/setup/roles/member.yaml` in the sandbox-sre repository
    - `PROJECT` file: https://github.com/mfrancisc/member-operator/blob/feature/ASC-224_SpaceRequest/PROJECT
    - `CSV` file: https://github.com/mfrancisc/member-operator/blob/feature/ASC-224_SpaceRequest/config/manifests/bases/member-operator.clusterserviceversion.yaml

    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/400
    - sandbox-sre: https://github.com/codeready-toolchain/sandbox-sre/pull/874
    - toolchain-cicd: https://github.com/codeready-toolchain/toolchain-cicd/pull/78


api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,86 @@
...
+// +k8s:openapi-gen=true
+type SpaceRequestStatus struct {
+
+	// TargetClusterURL The API URL of the cluster where Space is currently provisioned
+	// Can be empty if provisioning did not start or failed
+	// To be used to de-provision the NSTemplateSet if the Spec.TargetCluster is either changed or removed
+	// +optional
+	TargetClusterURL string `json:"targetClusterURL,omitempty"`
+
+	// The status of user accounts in the member clusters which belong to this MasterUserRecord
```
xcoulon, line 32: this comment seems wrong 🤔

api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,86 @@
...
+	// TierName is a required property introduced to retain the name of the tier
+	// for which this Space is provisioned
+	// If not set then the tier name will be set automatically
+	// +optional
+	TierName string `json:"tierName,omitempty"`
+
+	// TargetClusterLabels one or more labels that define a set of clusters
+	// where the Space can be provisioned.
+	// +optional
+	TargetClusterLabels map[string]string `json:"targetClusterLabels,omitempty"`
```
MatousJobanek, line 19: I would use just an array of strings instead of map
```suggestion
	TargetClusterLabels []string `json:"targetClusterLabels,omitempty"`
```
The labels at ToolchainClusters will be defined in the format:
```
toolchain.dev.openshift.com/cluster-label/workloads: ""
```
so we can have multiple labels defined for one CR. The actual value doesn't matter - what matters is the key of the label.
I took the inspiration from how k8s defines types of nodes: master, worker, or infra. They use the format:
```
node-role.kubernetes.io/worker: ''
```

So, if we don't care about the values, then I would just accept an array of strings that define the set of label keys it should match. 

cc @alexeykazakov & @rajivnathan as they were part of the discussion in the SpaceRequest document as well.
WDYT?

api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// SpaceRequestSpec defines the desired state of Space
+// +k8s:openapi-gen=true
+type SpaceRequestSpec struct {
+	// TierName is a required property introduced to retain the name of the tier
+	// for which this Space is provisioned
+	// If not set then the tier name will be set automatically
+	// +optional
+	TierName string `json:"tierName,omitempty"`
```
MatousJobanek, line 14: in the comment, you mentioned that the property is required, but you added the `// +optional` and `omitempty` markers, so it's not consistent. 
let's make it required for now until we have a nice way of handling default tiers for sub-spaces.
```suggestion
	// TierName is a required property introduced to retain the name of the tier
	// for which this Space is provisioned
	// If not set then the tier name will be set automatically
	TierName string `json:"tierName"`
```

api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,86 @@
...
+	TargetClusterLabels map[string]string `json:"targetClusterLabels,omitempty"`
+}
+
+// SpaceRequestStatus defines the observed state of Space
+// +k8s:openapi-gen=true
+type SpaceRequestStatus struct {
+
+	// TargetClusterURL The API URL of the cluster where Space is currently provisioned
+	// Can be empty if provisioning did not start or failed
+	// To be used to de-provision the NSTemplateSet if the Spec.TargetCluster is either changed or removed
```
MatousJobanek, line 28:
```suggestion
```
If the Spec.TargetCluster in the Space CR is changed, then the Space controller will take care of reprovisioning it to the right member cluster.
The URL is just for informative purposes for developers and controllers that are placed in member clusters.

api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,86 @@
...
+	// +optional
+	TargetClusterURL string `json:"targetClusterURL,omitempty"`
+
+	// The status of user accounts in the member clusters which belong to this MasterUserRecord
+	// +listType=atomic
+	NamespaceAccess []NamespaceAccess `json:"namespaceAccess,omitempty"`
```
MatousJobanek, line 34: let's mark it as optional
```suggestion
	// +optional
	NamespaceAccess []NamespaceAccess `json:"namespaceAccess,omitempty"`
```

api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Conditions is an array of current Master User Record conditions
+	// Supported condition types:
+	// Provisioning, UserAccountNotReady and Ready
```
MatousJobanek, line 38: could you please update this comment too?

api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,86 @@
...
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+
+// SpaceRequest is the Schema for the space request API
+// +k8s:openapi-gen=true
+// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:printcolumn:name="Tier",type="string",JSONPath=`.spec.tierName`
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
```
MatousJobanek, line 64: could we also add the TargetClusterURL?

api/v1alpha1/spacerequest_types.go
mfrancisc: done in : [04797c2d7449e7](https://github.com/codeready-toolchain/api/pull/333/commits/04797c2d7449e763fe1aa2f83a3058b5951df2c8)

api/v1alpha1/spacerequest_types.go
mfrancisc: done in : [04797c2d7449e7](https://github.com/codeready-toolchain/api/pull/333/commits/04797c2d7449e763fe1aa2f83a3058b5951df2c8)

api/v1alpha1/spacerequest_types.go
mfrancisc: right, copy pasting code has it's own drawbacks 🤦‍♂️ . thx!

fixed in: [04797c2d7449e7](https://github.com/codeready-toolchain/api/pull/333/commits/04797c2d7449e763fe1aa2f83a3058b5951df2c8)

api/v1alpha1/spacerequest_types.go
mfrancisc: My only point in favor of the map it's that sometimes it's easier to check if it has specific string keys.

For example checking if a cluster is workload type one would be something like :

```
if _, found := m["toolchain.dev.openshift.com/cluster-label/workloads"]; found {
    fmt.Println("is workload type cluster")
}
```

but I don't have strong opinions on map vs slice.
If the standard in k8s libraries is to use slices we should go with that I agree. They may have good reasons for it.

api/v1alpha1/spacerequest_types.go
mfrancisc: fixed in: [04797c2d7449e7](https://github.com/codeready-toolchain/api/pull/333/commits/04797c2d7449e763fe1aa2f83a3058b5951df2c8)

Now that I'm checking we have other CRDs with this same small issue. thx!

api/v1alpha1/spacerequest_types.go
mfrancisc: fixed in: [04797c2d7449e7](https://github.com/codeready-toolchain/api/pull/333/commits/04797c2d7449e763fe1aa2f83a3058b5951df2c8) 

thanks for additional details.

api/v1alpha1/spacerequest_types.go
mfrancisc: done in : [04797c2d7449e7](https://github.com/codeready-toolchain/api/pull/333/commits/04797c2d7449e763fe1aa2f83a3058b5951df2c8)

api/v1alpha1/spacerequest_types.go
alexeykazakov: Yeah... I like the array option better. And if we go with this option then let's rename `TargetClusterLabels` to `TargetClusterLabelKeys` to make it crystal clear that it's about the label keys. Not the entire label with key/value.

api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,87 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SpaceRequestSpec defines the desired state of Space
+// +k8s:openapi-gen=true
+type SpaceRequestSpec struct {
+	// TierName is a required property introduced to retain the name of the tier
+	// for which this Space is provisioned
+	// If not set then the tier name will be set automatically
+	TierName string `json:"tierName"`
```
alexeykazakov, line 13: It's required but "if not set then will be set automatically"? So is it required or optional? And will be set by what?

mfrancisc: yeah, that was a bit misleading, thx! Removed in [dcad81e788db5a](https://github.com/codeready-toolchain/api/pull/333/commits/dcad81e788db5a4c2a560b998f263082c577f558)

api/v1alpha1/spacerequest_types.go
mfrancisc: 👍  I've changed to slice of strings and renamed it, see : [dcad81e788db5a](https://github.com/codeready-toolchain/api/pull/333/commits/dcad81e788db5a4c2a560b998f263082c577f558) 

thanks for the feedback!

alexeykazakov (approved):
Looks good!

api/v1alpha1/spacerequest_types.go
xcoulon: Having a map of key/value pairs now instead of an array of keys could be helpful if in the future we also want to check the values, and it's doesn't make such a big difference in the model now. But changing it later will be trickier 🤷‍♂️

MatousJobanek: Good idea with the name `TargetClusterLabelKeys`

I see your point @xcoulon. However, since this is a "public-facing" API, I would love to have it simple - that's why I'm proposing the array. It will be other StoneSoup controllers who will create the CRs, not us. The fact that we store it as a label (key-value pair) is just our implementation detail. 
We could also name the parameter `TargetClusterTaints`, `TargetClusterRoles`, or `TargetClusterAPITypes`.

And if we would need to (for any reason) change it to a map in the future, then we can easily create a new parameter with a new name. That approach will be also easier for migration on the other controller's side

api/v1alpha1/spacerequest_types.go
```diff
@@ -0,0 +1,86 @@
...
+// SpaceRequestSpec defines the desired state of Space
+// +k8s:openapi-gen=true
+type SpaceRequestSpec struct {
+	// TierName is a required property introduced to retain the name of the tier
+	// for which this Space is provisioned.
+	TierName string `json:"tierName"`
+
+	// TargetClusterLabelKeys one or more label keys that define a set of clusters
+	// where the Space can be provisioned.
+	// +optional
```
MatousJobanek, line 16: Since it's been changed to a list, then we should also define the type, right?
```suggestion
	// +optional
	// +listType=atomic
```

MatousJobanek (approved):
Thanks for the additional changes. :+1:

api/v1alpha1/spacerequest_types.go
mfrancisc: good catch, thanks !

PR #335, opened 2023-01-09
add parentSpace field in Space CRD and minor comment fix

mfrancisc:
## Description
This PR adds `parentSpace` field to the `Space` CRD, covers the first acceptance criteria from https://issues.redhat.com/browse/ASC-247.

## Checks

2. Did `make generate` change anything in other projects? **yes**
host-operator

N/A

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/723

### DIFF
--- a/api/v1alpha1/space_types.go
+++ b/api/v1alpha1/space_types.go
@@ -51,8 +51,9 @@ type SpaceSpec struct {
 	// +optional
 	TierName string `json:"tierName,omitempty"`
 
-	// ParentSpace holds the name of the context (Space) from which this space was created (requested)
-	// using the SpaceRequest mechanism.
+	// ParentSpace holds the name of the context (Space) from which this space was created (requested),
+	// enabling hierarchy relationships between different Spaces.
+	//
 	// Keeping this association brings two main benefits:
 	// 1. SpaceBindings are inherited from the parent Space
 	// 2. Ability to easily monitor quota for the requested sub-spaces
--- a/api/v1alpha1/space_types.go
+++ b/api/v1alpha1/space_types.go
@@ -45,11 +45,19 @@ type SpaceSpec struct {
 	// +optional
 	TargetCluster string `json:"targetCluster,omitempty"`
 
-	// TierName is a required property introduced to retain the name of the tier
+	// TierName is introduced to retain the name of the tier
 	// for which this Space is provisioned
 	// If not set then the tier name will be set automatically
 	// +optional
 	TierName string `json:"tierName,omitempty"`
+
+	// ParentSpace holds the name of the context (Space) from which this space was created (requested)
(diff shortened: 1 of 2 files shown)


api/v1alpha1/space_types.go
```diff
@@ -45,11 +45,19 @@ type SpaceSpec struct {
...
 	TargetCluster string `json:"targetCluster,omitempty"`
 
-	// TierName is a required property introduced to retain the name of the tier
+	// TierName is introduced to retain the name of the tier
 	// for which this Space is provisioned
 	// If not set then the tier name will be set automatically
 	// +optional
 	TierName string `json:"tierName,omitempty"`
+
+	// ParentSpace holds the name of the context (Space) from which this space was created (requested)
```
xcoulon, line 54: given the description, should we call it `RequesterSpace` instead? (I know the Jira story already refers to `ParentSpace`), but I believe that k8s already uses the term `Requester` and I wonder if we could do the same here?

xcoulon: that being said, I reckon that `parentSpace` has a strong emphasis on the hierarchy relationship ;)

mfrancisc: Thank you both for review and detailed explanation.

Probably the comment should be more "broad" about how the hierarchy will be generated , since in the future will probably not be only by using SpaceRequest 🤔  

Maybe something like :

```
// ParentSpace holds the name of the context (Space) from which this space was created (requested),
// enabling hierarchy relationships between different Spaces.
//
// Keeping this association brings two main benefits:
// 1. SpaceBindings are inherited from the parent Space
// 2. Ability to easily monitor quota for the requested sub-spaces
```
WDYT ?

rajivnathan (approved):

api/v1alpha1/space_types.go
mfrancisc: changed in [48571b6](https://github.com/codeready-toolchain/api/pull/335/commits/48571b64a67a3202a21138209d8244afced68fab)

MatousJobanek (commented):
cool :+1: 
could you please add also the parentSpace label key here? 
https://github.com/codeready-toolchain/api/blob/48571b64a67a3202a21138209d8244afced68fab/api/v1alpha1/space_types.go#L7-L36

mfrancisc:
> could you please add also the parentSpace label key here?

done in [1cfba10](https://github.com/codeready-toolchain/api/pull/335/commits/1cfba10ff88199eb67263793274ef3492d0b7d55)

@MatousJobanek  thanks for pointing that out!

MatousJobanek (approved):
awesome, thanks :+1:

xcoulon (approved):

PR #338, opened 2023-02-06
feat: add list of provisioned namespaces into NSTemplateSet status

mfrancisc:
## Description
This PR adds the `NSTemplateSet.Status.ProvisionedNamespaces` field, which will contain all the namespace names provisioned by the NSTemplateSet.
The `Namespace.Type` field it's still WIP, at the moment only the first namespace in alphabetical order will have this field set to `default`, further work will be needed once we have better requirements about how many type of namespaces we need to support.

## Checks

2. Did `make generate` change anything in other projects ? **yes**
(host-operator, member-operator)

    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/407


api/v1alpha1/nstemplateset_types.go
```diff
@@ -85,6 +85,12 @@ type NSTemplateSetSpaceRole struct {
 // NSTemplateSetStatus defines the observed state of NSTemplateSet
 // +k8s:openapi-gen=true
 type NSTemplateSetStatus struct {
+	// ProvisionedNamespaces is a list of Namespaces that were provisioned by the NSTemplateSet.
+	// +optional
+	// +optional
+	// +listType=atomic
+	ProvisionedNamespaces []SpaceNamespace `json:"provisionedNamespaces,omitempty"`
```
MatousJobanek, line 92: Feel free to add it also to the Space CRD right away as part of this PR ;-)

mfrancisc: Ok, initially I've started down that path, but then I realized that I cannot add the logic to the space controller right away since I cannot pair both member , host with e2e tests. 

So you are suggesting to just add the new API change for space and then add the logic for nstemplateset and space in separate PRs right ?

MatousJobanek: never mind - I didn't notice that you already included the logic in the member operator. I was more thinking about changes in API only and then the operators separately. 
No issue, feel free to keep it as it is :wink:

mfrancisc: sorry, yeah I forgot to do it separately even this time. I promise I'll do it next time ! 😄 

In case a rename is required , I can do it quickly .

alexeykazakov (approved):

api/v1alpha1/nstemplateset_types.go
MatousJobanek: no worries, really :wink: :v: keep it as it is...

MatousJobanek (approved):

PR #344, opened 2023-02-17
feat: add SpaceRequestConfig object in ToolchainConfig CRD

mfrancisc:
## Description
Add a new `SpaceRequestConfig` object that will store all configuration values for the Space Request controller (which is still under development).

Specifically add a flag to enable/disable the SpaceRequest controller, so that it can be enabled only for StoneSoup environments and disabled by default in DevSandbox envs.

This PR just adds the new type , I'll open PRs to enable this flag in sandbox-sre as soon as this is merged to master.

Jira story: https://issues.redhat.com/browse/ASC-253  

## Checks

2. Did `make generate` change anything in other projects ? **yes**
host-operator

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/748


api/v1alpha1/toolchainconfig_types.go
```diff
@@ -77,6 +77,10 @@ type HostConfig struct {
 	// Keeps parameters necessary for configuring capacity limits
 	// +optional
 	CapacityThresholds CapacityThresholds `json:"capacityThresholds,omitempty"`
+
+	// Keeps parameters necessary for configuring SpaceRequest functionality
+	// +optional
+	SpaceRequest SpaceRequestConfig `json:"spaceRequest,omitempty"`
```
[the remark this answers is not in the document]
mfrancisc: I think it's a good idea yes! Let's see what others have to say about it...

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -445,6 +449,16 @@ type ToolchainConfigStatus struct {
...
 }
 
+// SpaceRequestConfig allows to configure , enable and disable the Space Request controller, which is in charge of provisioning Spaces on demand.
+// +k8s:openapi-gen=true
+type SpaceRequestConfig struct {
+
+	// Enabled specifies whether the SpaceRequest controller should start or not.
+	// This is specifically useful in order to enable/disable this functionality from configuration (e.g. disabled by default in Sandbox and enabled only for StoneSoup stage/prod ...).
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
alexeykazakov, line 459: Do we need it because if we just rely on kube RBAC because OLM will add `*` permissions to the aggregated "edit" role so all sandbox users will be able to create SpaceRequests in their namespaces?
I wonder if there is a way to disable including CRDs to the aggregated roles in OLM... Can you maybe ask the OLM folks about that before introducing this configuration?

mfrancisc: This will be used in host operator `main.go` file, once we will have the `SpaceRequest` Controller implemented, so that we can enable the controller reconcile logic only in StoneSoup environments, it will look something like this:

```
#main.go 

func main() {
# controllers setup logic ....

	if toolchainconfig.SpaceConfig.SpaceRequestEnabled {
		if err := (&spacerequest.Reconciler{
			Client: mgr.GetClient(),
			Scheme: mgr.GetScheme(),
		}).SetupWithManager(mgr); err != nil {
			setupLog.Error(err, "unable to create controller", "controller", "SpaceRequest")
			os.Exit(1)
		}
	}
	
# more setup here ....
}
```

As a sandbox user, right now, you are only able to list and get spacerequests in your namspace (tested this in my namespace `fmuntean-stage-dev` on sandbox staging environment) :
```
➜  kubectl auth can-i create spacerequests
no

➜  kubectl auth can-i list spacerequests
yes

➜ kubectl auth can-i get spacerequests
yes

➜ kubectl auth can-i update spacerequests
no

➜ kubectl auth can-i delete spacerequests
no
```
I don't know much about OLM, but I can have a look at how it creates/updates permissions, or I can check how to avoid deploying the CRD in case of sandbox environment. But apparently users do not have permissions to create/update/delete SpaceRequests.

alexeykazakov: Hm... I wonder why users with "edit" role (sandbox users) don't have permissions to create/update/delete SpaceRequest. I thought OLM would automatically add it to the aggregated "edit" and "admin" roles.
But it's good news!
I understand what this configuration is going to be used for.
My question is why do you need it?
Why not relying on RBAC?
If sandbox users can't create SpaceRequests in the first place why do we need to disable the controller?

mfrancisc: My bad, I'm not sure what happened with my staging account ( I'm not even able to register a new account apparently in staging, but let's discuss this separately).

So I've created a new account in prod, and yes you can create `SpaceRequest` resources actually!

```
➜  cat SpaceRequestTest.yaml
kind: SpaceRequest
metadata:
  name: spacerequest-sample
  namespace: fmuntea-devaccount-dev
spec:
  tierName: base
  targetClusterRoles:
    - tenant

➜  k apply -f SpaceRequestTest.yaml
spacerequest.toolchain.dev.openshift.com/spacerequest-sample created

➜  k get spacerequest -n fmuntea-devaccount-dev
NAME                  TIER   TARGETCLUSTERURL   READY   REASON
spacerequest-sample   base
``` 

So we do need to implement some sort of guardrail (IMO it's better to not rely on OLM aggregated roles anyway), so that users cannot create `SpaceRequest` (as it was thought initially ) , we can implement this using toolchainconfig as quick solution and open a new story for investing on how to do it with OLM or other ways (maybe not deploying the CRD at all if it's not supposed to be used). Or we can change the scope of the current Jira story and start investigating immediately if there's a different way to solve the problem.

api/v1alpha1/toolchainconfig_types.go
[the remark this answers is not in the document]
alexeykazakov: +1 to @rajivnathan's idea

alexeykazakov (approved):
Since it's impossible to avoid adding CRDs installed by OLM to aggregated roles let's proceed with introducing this configuration.
However please address the @rajivnathan's comment before merging :)

api/v1alpha1/toolchainconfig_types.go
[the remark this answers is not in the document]
mfrancisc: done see : https://github.com/codeready-toolchain/api/pull/344/commits/c35aaba802088d7254365f7906aa3020a0359b26

mfrancisc:
Thanks for reviewing ! Could you please take a look also at the related PR in host operator ?

PR #347, opened 2023-02-27
add notification environment in notification config

ranakan19:
## Description
A few sentences describing the overall goals of the pull request's commits.

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/754


api/v1alpha1/toolchainconfig_types.go
```diff
@@ -190,6 +190,10 @@ type NotificationsConfig struct {
 	// +optional
 	AdminEmail *string `json:"adminEmail,omitempty"`
 
+	// NotificationEnvironment is to distinguish the cluster environment being run in, eg: whether it is for sandbox or stonesoup
```
rajivnathan, line 193: Maybe add a sentence to explain it is used to switch between sets of notification templates?

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -194,6 +194,10 @@ type NotificationsConfig struct {
...
+	// +optional
+	NotificationEnvironment *string `json:"notificationEnvironment,omitempty"`
```
mfrancisc, line 199: I was wondering if it makes sense to use the already existing (and more generic) config option `toolchainconfig.Spec.host.environment` , so that it may be used in the future for other types of _switches_ . 

WDYT ?

ranakan19: I believe the existing config option environment is used to distinguish between prod and stage, so i didn't touch that and created a new one. Does renaming it help?

mfrancisc: You are right, we cannot reuse that. 

The name looks fine to me , I was wondering if we may have another use case for this environment switch (stonesoup|sandbox), other then differentiating the email templates, but I cannot think about another feature that may behave differently between sandbox and stonesoup right now. 

So I think we are fine for now 👍

api/v1alpha1/toolchainconfig_types.go
alexeykazakov: What if we rename it something more specific to templates? The environment terms is quite overloaded.
```suggestion
	// NotificationTemplateSetName defines the set of notification templates. Different Sandbox instances can use different notification templates. For example Dev Sandbox and AppStudio instances use different templates. By default the "sandbox" template set name is used.
```

api/v1alpha1/toolchainconfig_types.go
alexeykazakov, line 199: What if we rename it something more specific to templates? The environment terms is quite overloaded.
```suggestion
	// NotificationTemplateSetName defines the set of notification templates. Different Sandbox instances can use different notification templates. For example Dev Sandbox and AppStudio instances use different templates. By default the "sandbox" template set name is used.
	// +optional
	NotificationTemplateSetName *string `json:"notificationTemplateSetName,omitempty"`
```

alexeykazakov (approved):

MatousJobanek (approved):

rajivnathan (approved):

PR #355, opened 2023-04-03
Add captcha config

rajivnathan:
## Description
Adds configuration related to captcha verification.

Also adds UserSignup label key constant for captcha score.

## Checks

3. In case of **new** CRD, did you the following? **yes/no**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/784


alexeykazakov (approved):

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -383,6 +387,19 @@ type RegistrationServiceVerificationSecret struct {
 	AWSSecretAccessKey *string `json:"awsSecretAccessKey,omitempty"`
 }
 
+// CaptchaConfig defines any configuration related to captcha verification
+// +k8s:openapi-gen=true
+type CaptchaConfig struct {
+	// Enabled specifies whether the captcha verification feature is enabled or not
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
alexeykazakov, line 395: I'm thinking about a few options we need to support:
- No verification enabled at all (no phone verification, no captcha). That one is simple. We just disable both phone and captcha in our config.
- Captcha is enabled and if the score is below the threshold then it falls back to the phone verification. This is one is straight forward too. Both phone and captcha is enabled and the threshold is set.
- Only phone verification is required. This is what we currently have. No captcha is effectively enabled or taken into account. How we do it? Enable phone verification but disable the captcha?

And there is one more combination which is technically possible: Captcha is enabled but the phone verification is disabled. What happens in this case? Should we block all signups with the score below the threshold and only manual approval is possible for such signup requests?

Can we get it this clarified and documented in our API?

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -383,6 +387,19 @@ type RegistrationServiceVerificationSecret struct {
...
+
+	// ScoreThreshold defines the captcha assessment score threshold. A score equal to or above the threshold means the user is most likely human and
+	// can proceed signing up but a score below the threshold means the score is suspicious and further verification may be required.
+	// +optional
+	ScoreThreshold *string `json:"score,omitempty"`
```
mfrancisc, line 400: will this be converted to a numerical type at runtime for comparison or should we define it as float/int ?

rajivnathan: The scores we get are floats 0.0-1.0 but float is not a happy type for the api generation, it fails with an error. So it's a string that'll be converted at runtime and we'll have a default value in case of some invalid value. We don't have any floats anywhere in our api for this reason.

mfrancisc (approved):

api/v1alpha1/toolchainconfig_types.go
alexeykazakov: OK. I actually initially misread your PR. Thought you are introducing another filed **next to** the current `verification`. So basically I thought there will be `VerificationConfig` which would manage phone verification only and a new field `CaptchaConfig` to manage the captcha thing.

But now I see that you are introducing the captcha config as a sub-section of the `VerificationConfig`.
So, I think it's fine what you are suggesting:
- If `verification.enabled == false` then everything is disabled. Other fields/sub-sections do not matter.
- If it's `true` then the phone verification is enabled but the `verification.captcha` sub-section can be used to configure the captcha **on top of* the phone verification.
And there is no option to enable the captcha only without any phone verification at all.
We can revisit it later and introduce this option if we need it in the feature.

rajivnathan: Documented with be6ec0940bf3b6b06d8b79454c0d3cda7d3c8095

alexeykazakov (approved):

api/v1alpha1/usersignup_types.go
```diff
@@ -45,6 +45,9 @@ const (
 	// UserSignupUserPhoneHashLabelKey is used for the usersignup phone hash label key
 	UserSignupUserPhoneHashLabelKey = LabelKeyPrefix + "phone-hash"
 
+	// UserSignupCaptchaScoreLabelKey is set if captcha verification was used, and contains the last captcha assessment score for the user
+	UserSignupCaptchaScoreLabelKey = LabelKeyPrefix + "captcha-score"
```
MatousJobanek, line 49: wouldn't annotation make more sense? It doesn't have to be used as a label - we won't need to list UserSignup by the captcha score, right?

rajivnathan: Right. It should be an annotation. Good catch!

rajivnathan: Done with d7511b131a99949f5f93ac5ee0714d5a3aa4398f

MatousJobanek (approved):

PR #358, opened 2023-05-05
quality: Add Static Code Analysis checks

sawood14012:
## Description
Adding Static code analysis checks to run on each PR and merge

## Checks
1. Did you run `make generate` target? **yes/no**: **this PR does'nt affect any code**

2. Did `make generate` change anything in other projects (host-operator, member-operator)? **yes/no** : **this PR does'nt affect any code** 

3. In case of **new** CRD, did you the following? **yes/no**: **this PR does'nt affect any code** 

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/#
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/#


.github/workflows/analysis.yml
```diff
@@ -0,0 +1,18 @@
+name: Code Analysis
+on:
+  push:
+    branches:
+      - master
```
mfrancisc, line 5: If it's supposed to run on PRs as well as master branch.
```suggestion
  push:
    branches:
      - master
    tags-ignore:
      - '*.*'
  pull_request:
    branches:
      - master
```

sonar-project.properties
```diff
@@ -0,0 +1,2 @@
+sonar.projectKey=codeready-toolchain_api
+sonar.organization=codeready-toolchain
```
mfrancisc, line 2: Maybe we could add a description as well as Github link to the sourcecode (if not added automatically by the scanner )

.github/workflows/analysis.yml
sawood14012: yes i was thinking about this , only enabling pull_request wont do the trick as it requires the `SONAR_TOKEN` secret to send data but github doesnt supply secrets to forks i.e need to figure out a way

mfrancisc: I see we are doing a similar thing with the codecov token :https://github.com/codeready-toolchain/host-operator/blob/eadfaa0c2361240a79f7978932ae6a8daa91c49c/.github/workflows/ci-build.yml#L36
so I don't think it needs to supply the secret to the fork, but just use the secret when the PR is opened, but maybe I'm missing something.

And BTW I see sonar is already running for some PRs: https://github.com/codeready-toolchain/host-operator/pull/801
even if the job is not yet added to the github action workflow.

sawood14012: it will set the ci status to failed example : https://github.com/codeready-toolchain/host-operator/pull/800 so if we add this job as required we will be able to block PR's too

rajivnathan: Do we need it on master at all?

It's cool that we'll be able to block PRs eventually when it's stable and reliable but just to clarify we don't want to block PRs yet.

sawood14012: with automatic analysis we would not need it on master too we would just need the properties file (optional) @rajivnathan i will update this PR adding snyk check + properties file

MatousJobanek:
@sawood14012 is there any update on this PR?

PR #359, opened 2023-05-11
fix: replace "owner" label with "space" label

mfrancisc:
## Description
This PR adds `SpaceLabelKey` that will be set on the Namespace and namespace/cluster-scoped resources.
It will replace the `OwnerLabelKey` usage.

Jira: https://issues.redhat.com/browse/CRT-1776

## Checks

    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/438


api/v1alpha1/nstemplateset_types.go
```diff
@@ -6,6 +6,7 @@ import (
 
 const (
 	OwnerLabelKey       = LabelKeyPrefix + "owner"
+	SpaceLabelKey       = SpaceBindingSpaceLabelKey
```
mfrancisc, line 9: or maybe we could switch:
```
SpaceBindingSpaceLabelKey = SpaceLabelKey
``
?
🤔

mfrancisc: Yeah, I thought about that at first, but I didn't want to go over a bigger refactor that includes also renaming `SpaceBindingSpaceLabelKey` occurrences with `SpaceLabelKey` .  But I can do that of course.

rajivnathan: Makes sense, if it's too much refactoring then `SpaceBindingSpaceLabelKey = SpaceLabelKey` works too 👍

mfrancisc: The refactor by itself it's easy to do (with the fancy refactor IDE functionality 😄  ) , but I would have to open a PR for this in: regservice, host-operator, toolchain-common, toolchain-e2e and sandbox-sre. 
Maybe I can do this as an afterwards cleanup. WDYT ?

rajivnathan: I didn't look at all the places `SpaceBindingSpaceLabelKey` is used and opening multiple PRs is time consuming so if it requires opening PRs in all of those repos as a separate task then it's probably more work than is necessary for not much benefit so it's probably better to do as you suggested instead.

rajivnathan (approved):

api/v1alpha1/nstemplateset_types.go
alexeykazakov: Yeah.. I would stick with `SpaceBindingSpaceLabelKey = SpaceLabelKey` for now and maybe get rid of `SpaceBindingSpaceLabelKey` later.

alexeykazakov (approved):

api/v1alpha1/nstemplateset_types.go
mfrancisc: Changed in  https://github.com/codeready-toolchain/api/pull/359/commits/ebf567185f004fa6be2bbfd4bc501c0168bac26a

MatousJobanek (approved):

xcoulon (approved):

PR #362, opened 2023-05-26
feat: add environment field to memberoperator config  and version check reasons

mfrancisc:
## Description
Add Environment field in memberoperator config so that we can execute the commit SHA comparison only when we are in `prod`. 
Also add reason constants for: `DeploymentNotUpToDate` and `DeploymentIsUpToDate` that will be used in the Toolchainstatus conditions.

Jira: https://issues.redhat.com/browse/SANDBOX-53

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/814
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/444


MatousJobanek (approved):

api/v1alpha1/memberoperatorconfig_types.go
```diff
@@ -124,6 +132,26 @@ type ConsoleConfig struct {
 	RouteName *string `json:"routeName,omitempty"`
 }
 
+// GitHubConfig defines all parameters concerned with GitHub
+// +k8s:openapi-gen=true
+type GitHubConfig struct {
+	// Defines all secrets related to GitHub authentication/integration
+	// +optional
+	Secret GitHubSecret `json:"secret,omitempty"`
+}
```
MatousJobanek, line 141: why did you introduce another config instead of adding the `Secret GitHubSecret` to `MemberStatusConfig` and `ToolchainStatusConfig` - something like this:
```go
type MemberStatusConfig struct {
	// Defines the period between refreshes of the member status
	// +optional
	RefreshPeriod *string `json:"refreshPeriod,omitempty"`

	// Defines all secrets related to GitHub authentication/integration
	// +optional
	Secret GitHubSecret `json:"secret,omitempty"`
}
```
```go
type ToolchainStatusConfig struct {
	// ToolchainStatusRefreshTime specifies how often the ToolchainStatus should load and refresh the current hosted-toolchain status
	// +optional
	ToolchainStatusRefreshTime *string `json:"toolchainStatusRefreshTime,omitempty"`

	// Defines all secrets related to GitHub authentication/integration
	// +optional
	Secret GitHubSecret `json:"secret,omitempty"`
}
```

mfrancisc: Because I thought we might want to add more config related to Github, not just secrets. For example enabling and disabling the version check:

```
type GitHubConfig struct {
	// Defines all secrets related to GitHub authentication/integration
	// +optional
	Secret GitHubSecret `json:"secret,omitempty"`

        VersionCheck *bool `json:"versionCheck,omitempty"`
}
```

Or github org, repo names that now are constants in the operator source code. 
But I can remove them if we think it makes it more simple.

MatousJobanek: but the version check is supposed to be part of the `MemberStatusConfig`/`ToolchainStatusConfig` as this is the place where the check happens, right?
Let's keep the consistency and reference the `GitHubSecret` from the places where it is used

mfrancisc: It was just for keeping everything related to GitHub grouped under the same type, since `GitHubConfig` was part of `MemberStatusConfig/ToolchainStatusConfig`. But we can review this later if we need to add more config related to github.

Updated in [9601ce4](https://github.com/codeready-toolchain/api/pull/362/commits/9601ce4a7fc433700ce9494361c9d8774708111d)

PR #364, opened 2023-06-12
feat: monitor deployed versions of our sandbox operators

mfrancisc:
## Description
Add new Condition type to be used in toolchainstatus and memberoperator status. It will evidence if there is any drift in kind of deployed version of our operators compared to the source code version.

Jira: https://issues.redhat.com/browse/SANDBOX-53

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/819


api/v1alpha1/conditions.go
```diff
@@ -13,6 +13,9 @@ const (
 	// ConditionReady specifies that the resource is ready
 	ConditionReady ConditionType = "Ready"
 
+	// ConditionDeploymentVersion specifies if the deployment is up-to-date
+	ConditionDeploymentVersion ConditionType = "DeploymentVersion"
```
alexeykazakov, line 17: Shouldn't it be something like DeploymentVersionUpToDate?

Just DeploymentVersion type with True/False status could be a bit confusing

mfrancisc: Yeah, I was unsure about how to name it. I haven't added the `UpToDate` suffix since it's already present in the reasons https://github.com/codeready-toolchain/api/blob/c1f5bdd738974acd23cb18e590a5dbf43aec7b41/api/v1alpha1/toolchainstatus_types.go#L30-L31

I thought it may sound a bit redundant, having it both in the type and reason.

mfrancisc: Based on our discussion I've updated the name , plz see [43c5472](https://github.com/codeready-toolchain/api/pull/364/commits/43c5472af7424942ed15a35f1c337c35fa5c960b)

I'll update the other PRs (host and common) accordingly.

alexeykazakov (approved):
I guess you will add more reasons later. But so far looks good.

mfrancisc:
> I guess you will add more reasons later. But so far looks good.

Yeah , I'm checking , probably for the error cases I'll need to add some more reasons!

mfrancisc:
I've added `ToolchainStatusDeploymentVersionCheckGitHubErrorReason` to be used when there is an issue on GitHub API side.

rajivnathan (changes requested):
Can we discuss the need for a new condition type before going forward with the changes?

The ToolchainStatus and MemberStatus each have a single `Ready` condition and each have 'components' that tells us whether there's a problem. Each component also has a Ready condition and it bubbles up the chain up to the single ToolchainStatus condition. So for example, if the host operator deployment has a problem then the `type: Ready` condition should have `status: "False"` and a reason.

If the deployment version is outdated, then it means there's a problem with the deployment that needs to be addressed and I believe this fits into the `Ready` condition for the `hostOperator` component of the ToolchainStatus & the `memberOperator` component of MemberStatus.

Adding an additional condition type sort of complicates things a little because we now need to understand what this other type means and there can be many more types in the future but all we are trying to do here is check whether there is a problem that needs to be addressed.

I believe the reason for the new type was due to this comment https://github.com/codeready-toolchain/host-operator/pull/819#discussion_r1221704848 but I think we can handle this without adding a new type.

We can make hostOperator component Ready condition False and add a `revisionLastChecked` field to toolchainv1alpha1.HostOperatorStatus so that it is simple to understand if you look at the resource.

For example:
```
hostOperator:
      buildTimestamp: "2023-06-13T08:22:17Z"
      conditions:
      - lastTransitionTime: "2023-06-14T17:59:45Z"
        lastUpdatedTime: "2023-06-14T17:59:45Z"
        reason: DeploymentOutdated
        status: "False"
        type: Ready
      deploymentName: host-operator-controller-manager
      revision: aaa26d5754df75f814a2c5a7acc377e94ca17563
      version: 0.0.1
      revisionLastChecked: "2023-06-14T17:59:45Z"
```

We could also add a field for `revisionCheckEnabled` set to `true/false` if that's of interest.

WDYT?

mfrancisc:
@rajivnathan Initially I've implemented this by reusing the `Ready` condition type, but setting 2 conditions in the ToolchainStatus and MemberStatus, one was showing if the component is running correctly and the other was supposed to show if the component was up-to-date. That was initially the easiest path to implement.
But then after talking with @MatousJobanek I understood that we cannot have multiple conditions set with the same type, I guess this is a k8s convention or something internal only? 

We decided to have a separate condition because, in theory, it will make it simpler to understand and manage, since:
-  the `Ready` condition shows if the component is actually running, when this is set to `False` or `Unknown` we have to check and fix the problem "immediately", because it means that the deployment is probably not working/running 
- the new `DeploymentVersionUpToDate` condition instead provides insight on the version of the Deployment, just telling you that is up-to-date, is not up-to-date, or if this specific check is disabled. 

It can happen that (most cases) the `DeploymentVersionUpToDate` is set to `False/Unknown` but the `Ready` condition is `True`, in this case it means we can check exactly which commit was not deployed, and decide if it's something we must fix immediately or not (think about the case in which we just did a readme update or fixed a unit test or something else minor which doesn't represent an actual critical alarm but still needs to be fixed at some point)

In brief, from my point of view, having a separate condition makes it more clear. 

Also I'm not sure I understand what `revisionLastChecked: "2023-06-14T17:59:45Z"` means from your example. Could you elaborate more on how we will use this?

rajivnathan:
>It can happen that (most cases) the DeploymentVersionUpToDate is set to False/Unknown but the Ready condition is True, in this case it means we can check exactly which commit was not deployed, and decide if it's something we must fix immediately or not

IMO because we're on a continuous release cycle we should always want the operator to be up-to-date and if it's not then it's a problem that needs to be fixed ASAP regardless of the current impact because even if there was just a README change, there could be an important security fix that needs to go in immediately but could be blocked by this issue. I believe we introduced the cronjob to restart OLM because it's very important to keep the operators up-to-date and this check is an indication that there could be something wrong with the cronjob.

>Also I'm not sure I understand what revisionLastChecked: "2023-06-14T17:59:45Z" means from your example. Could you elaborate more on how we will use this?

`revisionLastChecked` would be a timestamp that's updated each time the revision is checked. I mentioned that since @MatousJobanek's concern in https://github.com/codeready-toolchain/host-operator/pull/819#discussion_r1221704848 was about having a way to see that the version check is actually being executed. If the `revisionLastChecked` timestamp is present and recent then we can be confident that the check is being executed. Adding a `revisionCheckEnabled` field would be further indication it is being executed, we'd expect it to be 'true' in prod and 'false' in other environments.

I'm not saying we should never have any other condition types, I just think this particular check is one of importance that we'd want to fix ASAP and having it as a separate condition type makes that less clear. But maybe others on the team feel different. 🤷

mfrancisc:
> IMO because we're on a continuous release cycle we should always want the operator to be up-to-date and if it's not then it's a problem that needs to be fixed ASAP regardless of the current impact because even if there was just a README change, there could be an important security fix that needs to go in immediately but could be blocked by this issue. I believe we introduced the cronjob to restart OLM because it's very important to keep the operators up-to-date and this check is an indication that there could be something wrong with the cronjob.

The difference I was trying to highlight is that when `Ready` condition is `False`, the issue becomes TOP priority for the team, when `DeploymentVersionUpToDate` is `False` it could be something to fix ASAP (which could be less critical).

 >revisionLastChecked would be a timestamp that's updated each time the revision is checked. I mentioned that since @MatousJobanek's concern in https://github.com/codeready-toolchain/host-operator/pull/819#discussion_r1221704848 was about having a way to see that the version check is actually being executed. If the revisionLastChecked timestamp is present and recent then we can be confident that the check is being executed. Adding a revisionCheckEnabled field would be further indication it is being executed, we'd expect it to be 'true' in prod and 'false' in other environments.
 
 Thanks for explaining it. I now understand your point. 
 Anyway, if there is an issue with the deployment (let's say it's not running) and also the version is not up-to-date, if we can only have 1 `Ready` condition we cannot notify both problems. Can we ? Which one we notify ?

rajivnathan:
>The difference I was trying to highlight is that when Ready condition is False, the issue becomes TOP priority for the team, when DeploymentVersionUpToDate is False it could be something to fix ASAP (which could be less critical).

Yeah, it would be nice to have severity. If we look at https://www.githubstatus.com/ for example, ToolchainStatus's main condition is like the overall system health ie. the "All Systems Operational" message at the top with the big green checkmark there. If there's a problem then it shouldn't be Ready (green). I'm not sure but I believe it can be green/orange/red with orange being operational but degraded and red being a more significant problem. All the subcomponents have their own status, which is similar to ToolchainStatus components and they can be green/orange/red. But they don't have multiple statuses, they have a single overall status. Because our conditions only have True/False we're limited to that but adding other conditions at the same level that aren't tied to the component status sounds like it breaks that relationship. How can the hostOperator have Ready=True and DeploymentVersionUpToDate=False at the same time? That would be like having Green and Orange at the same time. I think a HostOperatorStatus.DeploymentVersionUpToDate=True/False field would make more sense for exposing the status of that check specifically while keeping a single Ready condition for the component status.

>Anyway, if there is an issue with the deployment (let's say it's not running) and also the version is not up-to-date, if we can only have 1 Ready condition we cannot notify both problems. Can we ? Which one we notify ?

You can implement the code so that the deployment not running issue shows up as the reason ahead of the not up-to-date issue. And if we have the HostOperatorStatus.DeploymentVersionUpToDate=True/False field I mentioned above we'd still know there's an issue with the version too.

In the future we could consider creating a new type ( go type not ConditionType) called 'current-problems' or something that aggregates a list of current problems and includes severity information (eg. orange/red) and make it a subtype of components.


One problem we currently have is only getting notifications when the overall ToolchainStatus Ready condition changes, but we could change that so that we get notified if any component's Ready condition changes whether it's True/False status or the Reason changes.

MatousJobanek:
Thanks a lot for your feedback @rajivnathan :pray: 

Introducing another field `revisionLastChecked` to the operator status would technically work, but it's more or less almost the same as the last-update-time in Condition. In other words, we would try to expose almost the same information as we would do with the new Condition, but with the `revisionLastChecked` it would be incomplete information - we wouldn't be able to define the reason why the updated wasn't performed. Was it because the check or the secret is not configured? Or because of some other reason we can ignore? The admin would need to guess or know the implementation details of the check.

Looking at the concepts of conditions, showing when something was checked and updated for the last time with the results of the update is exactly what the Condition is supposed to be used for. Using condition gives us the flexibility of setting the Reason and Message to whatever we need without introducing another fields to the status. At the same time, we can reuse the same for all components:
* host-operator
* reg-service
* member-operator

> Adding an additional condition type sort of complicates things a little because we now need to understand what this other type means and there can be many more types in the future but all we are trying to do here is check whether there is a problem that needs to be addressed.

Having more conditions with different types should be always OK - we do that other CRDs - that's why it's defined as an array and not a single instance of the Condition per component.

I'm just trying to understand what the problem of introducing the new condition type actually is. 
1. Is it because we add a new condition type that is not derived from the Deployment?
2. is it because the Ready condition doesn't reflect the status of the new `DeploymentVersionUpToDate` condition as well (I haven't reviewed the last version of the PRs, but based on the discussion is suppose that this is what is there)

If it's the first one, then we could slightly change the structure of the host-operator status part. When you look at it and compare with reg-service or member statuses, the reg-service and members have sub-components
* for deployment
* healt-checks
* routes
* connections
* etc...

each of them has their own set of conditions. At the same time, the main component has it's own Ready condition reflecting statuses of the sub-components.
Host operator doesn't have that. Check for the host routes is placed in the [main ToolchainStatusStatus](https://github.com/codeready-toolchain/api/blob/8468f5843cf5df736714e4f52801f2f82c774c85/api/v1alpha1/toolchainstatus_types.go#L108-L110) instead of having it grouped together under the host component. That being said, we could move the `HostRoutes` into the host-operator part and create a new sub-component for the version check. Adding the same new version-check sub-component in `HostRegistrationServiceStatus` and `MemberStatusStatus` will be easy because it's already prepared for that - it will be just placed next to the already existing sub-components that I listed above.


If your concern is related to the second point - would it make more sense if the Ready condition of the host-operator would match the status of the `DeploymentVersionUpToDate` condition as well? In the same way as we do with other CRDs? In this way, we could easily keep adding new Condition types without risking that we could lose some information - the status of the Ready would always match the state of the other conditions - if any of them is False, then Ready is false as well. We could also copy the list of conditions from the Deployment CR.

Sorry for this long comment :innocent:

MatousJobanek (approved):
Looks great, thanks :+1:

PR #373, opened 2023-08-15
Introduced IdentityClaims embedded type to UserSignup.Spec

sbryzak:
## Description
This PR introduces a new IdentityClaims type to store as-is values of certain claims from the identity provider.

Fixes https://issues.redhat.com/browse/SANDBOX-324

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/850
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/471


api/v1alpha1/usersignup_types.go
```diff
@@ -195,6 +195,31 @@ type UserSignupSpec struct {
...
+	AccountID string `json:"accountID,omitempty"`
+
+	// GivenName contains the value of the 'given_name' claim
+	GivenName string `json:"givenName,omitempty"`
+
+	// FamilyName contains the value of the 'family_name' claim
+	FamilyName string `json:"familyName,omitempty"`
+
+	// Company contains the value of the 'company' claim
+	Company string `json:"company,omitempty"`
```
MatousJobanek, line 222: I would keep these separate. The reason is that the names stay in UserSignup CR only, but the IDs will be copied to MUR & UserAccount CRs

api/v1alpha1/usersignup_types.go
```diff
@@ -195,6 +195,31 @@ type UserSignupSpec struct {
...
+// +k8s:openapi-gen=true
+type IdentityClaimsEmbedded struct {
+	// Sub contains the value of the 'sub' claim
+	Sub string `json:"sub,omitempty"`
+
+	// UserID contains the value of the 'user_id' claim
+	UserID string `json:"userID,omitempty"`
+
+	// AccountID contains the value of the 'account_id' claim
+	AccountID string `json:"accountID,omitempty"`
```
MatousJobanek, line 213: how about the OriginalSub field?

api/v1alpha1/usersignup_types.go
alexeykazakov: We don't have to propagate everything from UserSignup.Spec.IdentityClaims to other resources like MUR and UserAccount. We still do it for selected fields like we do now. But having all sso token claims in one place (IdentityClaims) instead of spreading them as we do now is the whole point of this refactoring, right?

MatousJobanek: Well, when we talked last time, we discussed moving only the IDs into the separate struct, we didn't discuss doing that also for the other claims. 

The main problem that I wanted to solve was to have one place for all SSO IDs that need to go from the UserSignup into MUR & UserAccount. Right now, we have the IDs at multiple places and it's very chaotic. 
So in other words, I wanted to move the IDs from the annotations and spec into the separate field/struct.  In this way, as soon as we need to add another ID that need to be propagated down into the UserAccount, we would modify this single struct. Copying that over to MUR & UserAccount would be much easier. 

We don't have this problem with the names nor with the company field. we don't need to copy it anywhere, we don't have them stored in annotations either. The only thing we wanted to address wrt names was to keep it up-to-date with SSO.

So I'm slightly confused by what is the purpose of this PR. 
Do you want to keep all values from the SSO claims under a separate field. So why not moving the "username"  as well? And the email from the annotation? And if we moved everything under `identitytokenClaims` field, would be anything left in the spec?

Why not leaving the names where they are and move only the IDs to make the propagation easier?

```
spec:
  familyName:
  givenName: 
  company:
  username: mjobanek
  idClaims:
    sub:
    userID:
    accountID:
    originalSub:
```

api/v1alpha1/usersignup_types.go
sbryzak: added

alexeykazakov (approved):

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,48 @@ type UserSignupSpec struct {
 	// a new IdP provider client, and contains the user's "original-sub" claim
 	// +optional
 	OriginalSub string `json:"originalSub,omitempty"`
+
+	// IdentityClaims contains as-is claim values extracted from the user's access token
```
alexeykazakov, line 200:
```suggestion
        // +optional
	// IdentityClaims contains as-is claim values extracted from the user's access token
```

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,48 @@ type UserSignupSpec struct {
...
+	IdentityClaims IdentityClaimsEmbedded `json:"identityClaims,omitempty"`
+}
+
+// IdentityClaimsEmbedded is used to define a set of SSO claim values that we are interested in storing
+// +k8s:openapi-gen=true
+type IdentityClaimsEmbedded struct {
+
+	// PropagatedClaims
```
alexeykazakov, line 208: Not sure if we need it here...
```suggestion
	// PropagatedClaims
	// +optional
```

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,48 @@ type UserSignupSpec struct {
...
+	// +optional
+	FamilyName string `json:"familyName,omitempty"`
+
+	// Company contains the value of the 'company' claim
+	// +optional
+	Company string `json:"company,omitempty"`
+}
+
+type PropagatedClaims struct {
+	// UserID contains the value of the 'user_id' claim
```
alexeykazakov, line 231:
```suggestion
	// UserID contains the value of the 'user_id' claim
        // +optional
```

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,48 @@ type UserSignupSpec struct {
...
+	UserID string `json:"userID,omitempty"`
+
+	// AccountID contains the value of the 'account_id' claim
```
alexeykazakov, line 234:
```suggestion
	// AccountID contains the value of the 'account_id' claim
        // +optional
```

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,51 @@ type UserSignupSpec struct {
...
+	// +optional
+	IdentityClaims IdentityClaimsEmbedded `json:"identityClaims,omitempty"`
+}
+
+// IdentityClaimsEmbedded is used to define a set of SSO claim values that we are interested in storing
+// +k8s:openapi-gen=true
+type IdentityClaimsEmbedded struct {
+
+	// PropagatedClaims
+	PropagatedClaims PropagatedClaims `json:",inline"`
```
MatousJobanek, line 210: I thought that we wanted to embed the PropagatedClaims to make any future changes in the set of the fields easier.
```suggestion
	PropagatedClaims `json:",inline"`
```

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,51 @@ type UserSignupSpec struct {
...
+
+	// Sub contains the value of the 'sub' claim
+	Sub string `json:"sub,omitempty"`
```
MatousJobanek, line 213: Isn't sub the same as userID?

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,51 @@ type UserSignupSpec struct {
...
+
+	// Email contains the user's email address
+	Email string `json:"email,omitempty"`
```
MatousJobanek, line 216:
```suggestion
	// Email contains the user's email address
	// +optional
	Email string `json:"email,omitempty"`
```

api/v1alpha1/usersignup_types.go
alexeykazakov: No. This is one of the reason we are doing it. To get rid of the current confusion when we define the `sub` claim as `UserID` field. `user_id` and `sub` claim are different claims. In same cases they have the same value (when we use the sub claim override in some SSO clients) and some they are different (in console.redhat.com client for example).

api/v1alpha1/usersignup_types.go
alexeykazakov: Yes! Good catch.

api/v1alpha1/usersignup_types.go
alexeykazakov: Email is not actually optional claim.

MatousJobanek: good point. in that case:
```suggestion
	// Email contains the user's email address
	Email string `json:"email"`
```

api/v1alpha1/usersignup_types.go
MatousJobanek: I see now.
But shouldn't it be them moved into the PropagatedClaims section?

MatousJobanek: my point is, currently the claim is mapped to userID which is used for creating Identities in the member clusters - this should stay the same right?

api/v1alpha1/usersignup_types.go
sbryzak: fixed

api/v1alpha1/usersignup_types.go
sbryzak: fixed

api/v1alpha1/usersignup_types.go
sbryzak: @alexeykazakov informed me today that email was not optional

api/v1alpha1/usersignup_types.go
sbryzak: Apparently not always.

api/v1alpha1/usersignup_types.go
sbryzak: That is a bit nicer, good suggestion.

api/v1alpha1/usersignup_types.go
sbryzak: fixed

api/v1alpha1/usersignup_types.go
sbryzak: fixed

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,51 @@ type UserSignupSpec struct {
...
 	OriginalSub string `json:"originalSub,omitempty"`
+
+	// IdentityClaims contains as-is claim values extracted from the user's access token
+	// +optional
+	IdentityClaims IdentityClaimsEmbedded `json:"identityClaims,omitempty"`
+}
+
+// IdentityClaimsEmbedded is used to define a set of SSO claim values that we are interested in storing
+// +k8s:openapi-gen=true
+type IdentityClaimsEmbedded struct {
```
MatousJobanek, line 207: btw, how about the username - it's an SSO claim value as well, right?

api/v1alpha1/usersignup_types.go
sbryzak: Sorry, that was a response to your original comment "Isn't sub the same as userID"

api/v1alpha1/usersignup_types.go
sbryzak: you're right, we missed that.  Technically the claim is `preferred_username`, so I've added a property called `PreferredUsername`.

api/v1alpha1/usersignup_types.go
MatousJobanek: Ok, I understand now what you meant.
But still, shouldn't be the `Sub` field moved to `PropagatedClaims` so it gets into MUR & UserAccount CRs?

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,54 @@ type UserSignupSpec struct {
...
+
+	// PropagatedClaims
+	PropagatedClaims `json:",inline"`
+
+	// Sub contains the value of the 'sub' claim
+	Sub string `json:"sub,omitempty"`
+
+	// PreferredUsername contains the user's username
+	PreferredUsername string `json:"preferredUsername,omitempty"`
```
MatousJobanek, line 216: username should be required, right?
```suggestion
	PreferredUsername string `json:"preferredUsername"`
```

api/v1alpha1/usersignup_types.go
alexeykazakov: Yep. We need the sub claim to be propagated. And actually it looks like the only claim we do not need to propagate is `CompanyName`. We need email and full name to be propagated in all the places where we create email notifications...
Right?

sbryzak: I've moved sub to the propagated claims, and also made it required.

alexeykazakov: What about email and full name? What else we do propagate?

sbryzak: In MasterUserRecord we only have UserID and OriginalSub (plus the SSO annotations).  I don't see email and/or full name required elsewhere at present.

sbryzak: On the subject of notifications, I believe that we're only sending them from host operator currently and as far as I'm aware they're only generated off the UserSignup.

alexeykazakov: OK, good

api/v1alpha1/usersignup_types.go
```diff
@@ -196,6 +196,54 @@ type UserSignupSpec struct {
...
+type IdentityClaimsEmbedded struct {
+
+	// PropagatedClaims
+	PropagatedClaims `json:",inline"`
+
+	// PreferredUsername contains the user's username
+	PreferredUsername string `json:"preferredUsername"`
+
+	// Email contains the user's email address
+	Email string `json:"email"`
```
MatousJobanek, line 216: btw, we do propagate the email as well - I see that in MUR as well as in UserAccount as an annotation - this should be moved to `PropagatedClaims`

MatousJobanek (commented):
I just realized that the email is propagated down to UserAccount as well

PR #375, opened 2023-09-11
feat: add binding list to workspaces api

mfrancisc:
## Description
Add list of bindings to the current workspace API.

See jira: https://issues.redhat.com/browse/ASC-409

## Checks

2. Did `make generate` change anything in other projects (member-operator)? **yes**

    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/472


api/v1alpha1/workspace_types.go
```diff
@@ -27,6 +27,30 @@ type WorkspaceStatus struct {
 	// signing up. It is currently valid for this value to be empty.
 	// +optional
 	Type string `json:"type,omitempty"`
+
+	// AvailableRoles contains the roles for this tier. For example, "admin|contributor|maintainer".
+	// +optional
```
alexeykazakov, line 32:
```suggestion
	// +listType=atomic
	// +optional
```

api/v1alpha1/workspace_types.go
```diff
@@ -27,6 +27,30 @@ type WorkspaceStatus struct {
...
+	AvailableRoles []string `json:"availableRoles,omitempty"`
+
+	// Bindings enumerates the permissions that have been granted to users within the current workspace, and actions that workspace admin users can apply to those permissions.
+	// +optional
```
alexeykazakov, line 36:
```suggestion
	// +listType=atomic
	// +optional
```

api/v1alpha1/workspace_types.go
```diff
@@ -27,6 +27,30 @@ type WorkspaceStatus struct {
 	// signing up. It is currently valid for this value to be empty.
 	// +optional
 	Type string `json:"type,omitempty"`
+
+	// AvailableRoles contains the roles for this tier. For example, "admin|contributor|maintainer".
+	// +optional
+	AvailableRoles []string `json:"availableRoles,omitempty"`
+
+	// Bindings enumerates the permissions that have been granted to users within the current workspace, and actions that workspace admin users can apply to those permissions.
```
alexeykazakov, line 35:
```suggestion
	// Bindings enumerates the permissions that have been granted to users within the current workspace, and actions that can be applied to those permissions.
```

api/v1alpha1/workspace_types.go
```diff
@@ -27,6 +27,30 @@ type WorkspaceStatus struct {
...
+	MasterUserRecord string `json:"masterUserRecord,omitempty"`
+
+	// Role is the role of the user in the current workspace. For example "admin" for the user that has all permissions on the current workspace.
+	Role string `json:"role,omitempty"`
+
+	// AvailableActions is a list of actions that can be performed on the binding.
+	// Available values:
+	// - "update" when the role in the current binding can be changed
+	// - "delete" when the current binding can be deleted
+	// - "override" when the current binding can be overwritten by a new binding containing same MasterUserRecord but different role
```
alexeykazakov, line 51:
```suggestion
	// - "override" when the current binding can be overwritten by a new binding containing the same MasterUserRecord but a different role
	// +listType=atomic
```

alexeykazakov (approved):
Looks good! I left a few comments though.

api/v1alpha1/workspace_types.go
```diff
@@ -27,6 +27,30 @@ type WorkspaceStatus struct {
...
+	AvailableRoles []string `json:"availableRoles,omitempty"`
+
+	// Bindings enumerates the permissions that have been granted to users within the current workspace, and actions that workspace admin users can apply to those permissions.
+	// +optional
+	Bindings []Binding `json:"bindings,omitempty"`
+}
+
+type Binding struct {
+	// MasterUserRecord is the name of the user that has access to the workspace.
+	MasterUserRecord string `json:"masterUserRecord,omitempty"`
```
rajivnathan, line 42: Worth adding a sentence here about it being immutable (via webhook)?

api/v1alpha1/workspace_types.go
rajivnathan, line 51: I found the override text a bit confusing tbh. I think it was more clear in the doc so maybe something like this would be better? WDYT?
```suggestion
	// - "override" when the current binding is inherited from a parent workspace, it cannot be updated but it can be overridden by creating a new binding containing the same MasterUserRecord but different role in the subworkspace.
```

api/v1alpha1/workspace_types.go
mfrancisc: Thanks ! I've missed all those listTypes 🤦‍♂️

api/v1alpha1/workspace_types.go
mfrancisc: Makes sense, mentioning why it cannot be updated indeed. Thanks!

mfrancisc: plz see https://github.com/codeready-toolchain/api/pull/375/commits/84f32f2478cecfd12bedc147b8e65245b3d045ed

api/v1alpha1/workspace_types.go
mfrancisc: Good point, plz check https://github.com/codeready-toolchain/api/pull/375/commits/84f32f2478cecfd12bedc147b8e65245b3d045ed

mfrancisc:
@alexeykazakov @rajivnathan thanks for your review! I've addressed your comments, PTAL when you have time.

alexeykazakov (approved):

rajivnathan (approved):
Looks good!

PR #382, opened 2023-10-12
fix: binding request field should not be returned when empty

mfrancisc:
## Description
Fix the issue with binding request field in workspace api being returned even when empty. 

As the [enconding/json](https://pkg.go.dev/encoding/json#Marshal) package says, "omitempty" works only with nil pointers when dealing with structs:

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

## Checks

2. Did `make generate` change anything in other projects? **no**

    - registration-service: https://github.com/codeready-toolchain/registration-service/pull/358
    - toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/818


alexeykazakov (approved):

alexeykazakov:
Good catch!

rajivnathan (approved):

MatousJobanek (commented):
:+1:

MatousJobanek (approved):

PR #384, opened 2023-10-18
Add HomeSpace status field

baijum:
## Description

This field is going to be used by the proxy when no workspace context is provided.

Ref. Ref. https://issues.redhat.com/browse/SANDBOX-425

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/892

### DIFF
--- a/api/v1alpha1/usersignup_types.go
+++ b/api/v1alpha1/usersignup_types.go
@@ -266,7 +266,8 @@ type UserSignupStatus struct {
 	// +optional
 	CompliantUsername string `json:"compliantUsername,omitempty"`
 
-	// HomeSpace represents the default home space for the user.
+	// HomeSpace is the name of the Space that is created for the user
+	// immediately after their account is approved.
 	// This is used by the proxy when no workspace context is provided.
 	// +optional
 	HomeSpace string `json:"homeSpace,omitempty"`
--- a/api/v1alpha1/usersignup_types.go
+++ b/api/v1alpha1/usersignup_types.go
@@ -265,6 +265,11 @@ type UserSignupStatus struct {
 	// CompliantUsername is used to store the transformed, DNS-1123 compliant username
 	// +optional
 	CompliantUsername string `json:"compliantUsername,omitempty"`
+
+	// HomeSpace represents the default home space for the user.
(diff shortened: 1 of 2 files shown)


api/v1alpha1/usersignup_types.go
```diff
@@ -265,6 +265,11 @@ type UserSignupStatus struct {
 	// CompliantUsername is used to store the transformed, DNS-1123 compliant username
 	// +optional
 	CompliantUsername string `json:"compliantUsername,omitempty"`
+
+	// HomeSpace represents the default home space for the user.
```
mfrancisc, line 269: minor thing - maybe we could specify that this field will contain the name of the Space that is created for the user immediately after their account is approved. WDYT?

mfrancisc (commented):
Looks good 👍 

I have one minor comment.

api/v1alpha1/usersignup_types.go
baijum: Thanks for the review!
Done. ab5bad292c4da077bab1ed3aad1bbc564957ec06

mfrancisc: Thanks! 

It seems like the the generated.openapi.go file still has the old description for some reason : https://github.com/codeready-toolchain/api/pull/384/files#diff-72ca28827e299dac63194a397a04873f7f4d9fe3a216136c97a81c5e561d92acR4931

baijum: Ah. I missed copying that from the generated location. I had to copy it like this:
`cp github.com/codeready-toolchain/api/api/v1alpha1/zz_generated.openapi.go api/v1alpha1/`

Fixed here: bf58427eb318ca19f0a19fa26c3223b2a6d3986e

mfrancisc: you can avoid the copy if you clone the projects into your `GOPATH` location for example:
`GOPATH="~/go"`
and projects should be located at:
`
~/go/src/github.com/codeready-toolchain/api
`

mfrancisc (approved):
Thanks for addressing my comments!

rajivnathan (approved):

PR #385, opened 2023-11-01
Create webhook secret

rajivnathan:
## Description
Adds another webhook configuration field to the memberoperatorconfig api.

## Checks

3. In case of **new** CRD, did you the following? **n/a**

    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/491


alexeykazakov (approved):

MatousJobanek (approved):

api/v1alpha1/memberoperatorconfig_types.go
```diff
@@ -158,6 +158,22 @@ type WebhookConfig struct {
 	// Defines the flag that determines whether to deploy the Webhook
 	// +optional
 	Deploy *bool `json:"deploy,omitempty"`
+
+	// Defines all secrets related to webhook configuration
+	// +optional
+	Secret WebhookSecret `json:"secret,omitempty"`
```
mfrancisc, line 164: For `omitempty` to work with structs, the variable has to be a pointer
```suggestion
	Secret *WebhookSecret `json:"secret,omitempty"`
```

api/v1alpha1/memberoperatorconfig_types.go
```diff
@@ -158,6 +158,22 @@ type WebhookConfig struct {
...
+// WebhookSecret defines all secrets related to webhook configuration
+// +k8s:openapi-gen=true
+type WebhookSecret struct {
+	// The reference to the secret that is expected to contain the keys below
+	// +optional
+	ToolchainSecret `json:",inline"`
+
+	// The key in the secret values map that contains a comma-separated list of SSH keys
+	// +optional
+	VirtualMachineAccessKey *string `json:"virtualMachineAccessKey,omitempty"`
```
mfrancisc, line 176: just curious - why a pointer to string?

mfrancisc (commented):
Looks good 👍 

I have few minor comments.

api/v1alpha1/memberoperatorconfig_types.go
MatousJobanek: I think that it should be fine - we use that in this way at other several places. 
When you look at ToolchainConfig, then the optional struct is used without having pointer all over the place

api/v1alpha1/memberoperatorconfig_types.go
MatousJobanek: we need to differentiate between 3 values:
1. value is not provided (pointer is nil)
2. value is empty ""
3. value is non-empty

we use the pointer for the actual values in ToolchainConfig & MemberConfig CRDs to differentiate between 1. and 2. situations

api/v1alpha1/memberoperatorconfig_types.go
mfrancisc: Right , it's a minor thing. The only thing is that those fields will show up in the CR even when empty.

api/v1alpha1/memberoperatorconfig_types.go
mfrancisc: makes sense thanks for explaining that!

api/v1alpha1/memberoperatorconfig_types.go
mfrancisc: I'm not sure about ToolchainConfig and MUR.Spec.PropagatedClaims. But I've looked at the `memberoperatorconfig` for example and you can notice the empty fields:
```
....
  spec:
    console: {}
    toolchainCluster: {}
    webhook: {}
```
the `omitempty` tag doesn't seem to work there.
Again not a big issue.

rajivnathan: It's not a big issue but it's a good catch. I think the pattern is for the top-level config to not use pointers but for the nested config to use pointers so that they don't show if empty. WebhookSecret is a nested config so I went ahead and changed it to use a pointer.

MatousJobanek: That's interesting... when you get the MemberOperatorConfig, then the empty values are visible, but when you get Toolchainconfig, then the same fields for the same struct (in `members` section) are hidden. 
So it works for some CRDs, but for some it doesn't :man_shrugging:

mfrancisc: yeah that's weird indeed 😕

PR #390, opened 2023-11-28
add automatic verification threshold field

mfrancisc:
## Description
Add a new field for configuring the automatic verification threshold.

Jira: https://issues.redhat.com/browse/SANDBOX-495

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/942

### DIFF
--- a/api/v1alpha1/toolchainconfig_types.go
+++ b/api/v1alpha1/toolchainconfig_types.go
@@ -446,6 +446,10 @@ type CaptchaConfig struct {
 	// +optional
 	AutomaticVerificationThreshold *string `json:"automaticVerificationThreshold,omitempty"`
 
+	// AllowLowScoreReactivation specifies whether the reactivation for users with low captcha score (below AutomaticVerificationThreshold) is enabled without the need for manual approval.
+	// +optional
+	AllowLowScoreReactivation *bool `json:"allowLowScoreReactivation,omitempty"`
+
 	// SiteKey defines the recaptcha site key to use when making recaptcha requests. There can be different ones for different environments. eg. dev, stage, prod
 	// +optional
 	SiteKey *string `json:"siteKey,omitempty"`
--- a/api/v1alpha1/toolchainconfig_types.go
+++ b/api/v1alpha1/toolchainconfig_types.go
@@ -441,12 +441,12 @@ type CaptchaConfig struct {
 	// +optional
 	ScoreThreshold *string `json:"scoreThreshold,omitempty"`
 
-	// AutomaticVerificationThreshold defines the lowest captcha score for automatic approval to be enabled.
-	// If the captcha score is below this threshold then manual approval is required in order for the user to be provisioned.
+	// RequiredScore defines the lowest captcha score, below this score the user cannot proceed with the signup process at all.
+	// Users with captcha score lower than the required one can still be approved manually.
 	// +optional
-	AutomaticVerificationThreshold *string `json:"automaticVerificationThreshold,omitempty"`
+	RequiredScore *string `json:"requiredScore,omitempty"`
 
-	// AllowLowScoreReactivation specifies whether the reactivation for users with low captcha score (below AutomaticVerificationThreshold) is enabled without the need for manual approval.
+	// AllowLowScoreReactivation specifies whether the reactivation for users with low captcha score (below the RequiredScore) is enabled without the need for manual approval.
 	// +optional
 	AllowLowScoreReactivation *bool `json:"allowLowScoreReactivation,omitempty"`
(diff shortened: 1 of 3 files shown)


alexeykazakov (approved):

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -441,6 +441,11 @@ type CaptchaConfig struct {
 	// +optional
 	ScoreThreshold *string `json:"scoreThreshold,omitempty"`
 
+	// AutomaticVerificationThreshold defines the lowest captcha score for automatic approval to be enabled.
+	// If the captcha score is below this threshold then manual approval is required in order for the user to be provisioned.
+	// +optional
+	AutomaticVerificationThreshold *string `json:"automaticVerificationThreshold,omitempty"`
```
MatousJobanek, line 447: I would maybe call it 
`MinimalRequiredScoreThreshold` or just `MinimalRequiredScore` because if the user gets a low score, then we don't let the user even try to complete the phone verification. In other words, it's not only about automatic approval.

mfrancisc: Thanks for your suggestion @MatousJobanek !

I was unsure about the naming (as usual 😅 ) , I thought about something using the words `minimal required` . But TBH to me it doesn't explain the purpose of the configuration. It may not be clear what happens if the score is lower then the config. While the "automatic verification"  implies that the user can proceed with the verification (phone, email, 2fa ... ) without an admin approval/intervention. Maybe autonomous or self verification would be better words 🤔 

But again I don't have a strong opinion on the naming, whatever makes more sense to people 🤷‍♂️

alexeykazakov: > In other words, it's not only about automatic approval - that's what the ScoreThreshold field is used for actually

Well... automatic approval is configured via `AutomaticApprovalConfig`. Not via ScoreThrershold. Even if PhoneVerification is required it's still auto approval if it's enabled in AutomaticApprovalConfiguration. Vs. a manual approval by an admin.
We could move this new field to AutomaticApprovalConfig and call it something like `MinimalRequiredCaptchaScore` or keep it under the Captcha configuration and call it something like `MinimalRequiredScoreForAutoApproval` if it's not too long :)
Otherwise I would probably prefer @mfrancisc's variant.

MatousJobanek: I'm idiot, I read "AutomaticApprovalThreshold" instead of what is there "AutomaticVerificationThreshold". I have some issues with my eyes in the last two days, so it's sometimes hard to focus on the words on the display, or I just maybe need a coffee. 

I got maybe also confused by the comment talking about "automatic approval:
```
// AutomaticVerificationThreshold defines the lowest captcha score for automatic approval to be enabled.
// If the captcha score is below this threshold then manual approval is required in order for the user to be provisioned.
```
Because as Alexey mentioned, this threshold doesn't have anything to do with the automatic/manual approval, it's about proceeding with the signup process.
I would at least drop the "Automatic" prefix to not confuse more people. 

so other ideas:
* SignupProcessThreshold
* RequiredScore
* HardStopThreshold
* GatekeeperScore
* SignupConfidenceLimit

Or just go with your name and update the comments so it doesn't talk about the automatic/manual approval, but rather that the user cannot proceed with the signup process at all

mfrancisc: No worries, and yeah the comment it's unclear.

Thanks for your suggestions, I like both `SignupProcessThreshold` and `RequiredScore`. 
Maybe we can go with `RequiredScore` since it's also closer to what @alexeykazakov is suggesting ?

mfrancisc: it will be `config.Spec.Host.RegistrationService.Verification.Captcha.RequiredScore`

MatousJobanek: sounds good :+1:

mfrancisc: renamed in https://github.com/codeready-toolchain/api/pull/390/commits/9247ee8fe6b173cd1a31f719240624775ae31b4a

I've also added a new field in https://github.com/codeready-toolchain/api/pull/390/commits/baa8ab5baf236208e35f90e5ecfe2f8838ba0cc5 as per slack discussion.

MatousJobanek (approved):
Nice, thanks :+1:

mfrancisc:
@alexeykazakov could you PTAL when you have some time?

alexeykazakov (approved):
Looks good to me!

PR #397, opened 2024-01-15
KSPACE-28 - Defines the SpaceProvisionerConfig CRD

metlos:
## Description
This adds the `SpaceProvisionerConfig` CRD.

Related issues:
https://issues.redhat.com/browse/KSPACE-4 and https://issues.redhat.com/browse/KSPACE-28

## Checks

See the PR links below

    - `resources/setup/roles/host.yaml` in `sandbox-sre` repository

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/959
    - sandbox-sre: https://github.com/codeready-toolchain/sandbox-sre/pull/1430
    - toolchain-cicd: https://github.com/codeready-toolchain/toolchain-cicd/pull/98


api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -0,0 +1,79 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+const (
+	SpaceProvisionerConfigValidConditionType             = "Valid"
```
MatousJobanek, line 6: In most of our CRDs we use the `Ready` condition type.
https://github.com/codeready-toolchain/api/blob/d5c7803671c1e0603b63761c07748d5910461962/api/v1alpha1/conditions.go#L14
 I believe that we could use it for this CRD too instead of bringing a new type

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -0,0 +1,79 @@
...
+	SpaceProvisionerConfigValidReason                    = "AllChecksPassed"
+)
+
+// +k8s:openapi-gen=true
+type SpaceProvisionerConfigSpec struct {
+	// ToolchainCluster is the name of the ToolchainCluster CR of the member cluster that this config is for.
+	ToolchainCluster string `json:"toolchainCluster"`
+
+	// Enabled specifies whether the member cluster is enabled (and therefore can hold spaces) or not.
+	Enabled bool `json:"enabled"`
```
MatousJobanek, line 17: this field can be optional - by default `false` == disabled

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// CapacityThresholds specifies the max capacities allowed in this provisioner
+	CapacityThresholds SpaceProvisionerCapacityThresholds `json:"capacityThresholds"`
```
MatousJobanek, line 20: let's make this optional as well including the actual values. By default, if no thresholds provided then it means that there are no limits

api/v1alpha1/spaceprovisionerconfig_types.go
metlos: With the explicit default, Kubernetes actually writes the value to etcd. So then when you list the objects, you can see that the value is `false` even if you didn't specify it. Without the default, kubectl/oc renders it empty in the listing which IMHO is a little confusing.

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// PlacementRoles is the list of roles, or flavors, that the provisioner possesses that influence
+	// the space scheduling decisions.
+	// +optional
+	PlacementRoles *[]string `json:"placementRoles,omitempty"`
```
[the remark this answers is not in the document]
metlos: From golang's perspective, it of course would be better if this weren't a pointer. I did this because of the fact that the default value for array is an empty array and therefore, even if you don't specify it it ends up as an empty array in `oc get ... -o yaml`. When I think about it again, I guess that actually fine :)

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// +k8s:openapi-gen=true
+type SpaceProvisionerConfigStatus struct {
+	// Conditions describes the state of the configuration (its validity).
+	// The only known condition type is "Ready".
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
MatousJobanek, line 58: btw, we have our own Condition type https://github.com/codeready-toolchain/api/blob/d5c7803671c1e0603b63761c07748d5910461962/api/v1alpha1/conditions.go#L28-L45 that we use in most of our CRDs, and we have a bunch of helper functions for that https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/condition/condition.go

I'm aware of this Condition from apimachinery. We can eventually replace most of the fields using the k8s conditon, but let's do it later as a separate task and let's stay what we have there for now.

Also, you can define the merge strategy (not sure what is the default one) and the merge key:
```suggestion
	// +optional
	// +patchMergeKey=type
	// +patchStrategy=merge
	// +listType=map
	// +listMapKey=type
	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -0,0 +1,92 @@
...
+	ToolchainCluster string `json:"toolchainCluster"`
+
+	// Enabled specifies whether the member cluster is enabled (and therefore can hold spaces) or not.
+	// +optional
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled"`
+
+	// CapacityThresholds specifies the max capacities allowed in this provisioner
+	// +optional
+	CapacityThresholds SpaceProvisionerCapacityThresholds `json:"capacityThresholds"`
```
MatousJobanek, line 27:
```suggestion
	CapacityThresholds SpaceProvisionerCapacityThresholds `json:"capacityThresholds,omitempty"`
```

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -0,0 +1,92 @@
...
+	// The only known condition type is "Ready".
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// SpaceProvisionerConfig is the configuration of space provisioning in the member clusters.
+//
+// Note that these objects are currently NOT used anywhere.
```
MatousJobanek, line 63: did you add this "note" intentionally?

MatousJobanek (commented):
btw, we also need to update the add-cluster.sh script 
https://github.com/codeready-toolchain/toolchain-cicd/blob/9313a2f795ec3fe10f6219225d951cec759910c2/scripts/add-cluster.sh#L61-L73

api/v1alpha1/spaceprovisionerconfig_types.go
metlos: Yes, because it is going to take a little bit of time before these will be integrated into the cluster provisioning workflow. Now, we're just creating the CRD, next is the controller that doesn't really do anything apart from checking the validity of the CRs. So it will be a couple of weeks at least between this being in the codebase and being actively used for the intended purpose.

api/v1alpha1/spaceprovisionerconfig_types.go
metlos: E.g. https://play.golang.com/p/JTbLwV_WmBY

MatousJobanek: good point - I didn't realize it before, but it makes sense :-)

api/v1alpha1/spaceprovisionerconfig_types.go
MatousJobanek: let's remove it, there is technically no one else who would rely on the api except for DevSandbox and Konflux, and they both know what is going on. 
I understand the intention and theoretically it's correct approach, but it's not worth maintaining such a comments during the time.

MatousJobanek (approved):
Looks good. :+1: 
Thanks for adding the extra changes and also providing the additional information :pray: :rocket:

PR #404, opened 2024-03-12
Sandbox 554

sbryzak:
## Description
Introduce a new property to the UserSignup.Status which stores the timestamp of the user's deactivation.  Related PR: https://github.com/codeready-toolchain/registration-service/pull/411

## Checks

3. In case of **new** CRD, did you the following? **n/a**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/988


api/v1alpha1/usersignup_types.go
```diff
@@ -247,6 +247,12 @@ type UserSignupStatus struct {
 	// This is used by the proxy when no workspace context is provided.
 	// +optional
 	HomeSpace string `json:"homeSpace,omitempty"`
+
+	// DeactivationTimestamp is the calculated timestamp after which the user's account will be deactivated, typically
+	// after the expiry of their trial and based on the term specific by their UserTier.  This property may be used as
+	// a convenience to determine the amount of time an account has left before deactivation, without requiring a separate
+	// lookup for the UserTier and subsequent calculation.  It is managed by the Deactivation controller in the host operator.
+	DeactivationTimestamp string `json:"deactivationTimestamp,omitempty"`
```
xcoulon, line 255: shouldn't it be stored in one of the `status.conditions`, instead? We already add a status condition when a user is activated or banned, so why using something different for deactivations?

xcoulon: ah never mind, this is the _expected_ time of deactivation, not the actual one (ie, after the user was deactivated).

xcoulon: so this field will need to be updated when a user is moved to another tier.

MatousJobanek: Let's maybe rename it to `ExpectedTimeOfDeactivation` or something like that to make it clear.

MatousJobanek: Also, could you please explain why you decided to use `string`? what is the expected format/content?
Why don't you use `metav1.Time` as we use in conditions?

alexeykazakov: I agree with @MatousJobanek. Something like `ExpectedTimeOfDeactivation` with `metav1.Time` type would look better.

sbryzak: I've renamed it to `ScheduledDeactivationTimestamp`, and also modified the type to use `metav1.Time` instead of a string to bring it inline with the rest of the API (before I saw these comments, so at least we were all on the same page ;).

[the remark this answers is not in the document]
sbryzak: Yes it should be, fixed.

api/v1alpha1/usersignup_types.go
```diff
@@ -247,6 +247,13 @@ type UserSignupStatus struct {
 	// This is used by the proxy when no workspace context is provided.
 	// +optional
 	HomeSpace string `json:"homeSpace,omitempty"`
+
+	// ScheduledDeactivationTimestamp is the calculated timestamp after which the user's account will be deactivated, typically
+	// after the expiry of their trial and based on the term specific by their UserTier.  This property may be used as
+	// a convenience to determine the amount of time an account has left before deactivation, without requiring a separate
+	// lookup for the UserTier and subsequent calculation.  It is managed by the Deactivation controller in the host operator.
+	// +optional
+	ScheduledDeactivationTimestamp metav1.Time `json:"deactivationTimestamp,omitempty"`
```
alexeykazakov, line 256: Why not:
```suggestion
	ScheduledDeactivationTimestamp metav1.Time `json:"scheduledDeactivationTimestamp,omitempty"`
```
?

sbryzak: Thanks for catching that, fixed.

alexeykazakov (approved):
:+1:

PR #410, opened 2024-03-22
add PublicViewerConfig in ToolchainConfig

filariow:
## Description
JIRA: https://issues.redhat.com/browse/ASC-492

## Checks

- host-operator

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/998


api/v1alpha1/toolchainconfig_types.go
```diff
@@ -32,6 +32,30 @@ type ToolchainConfigSpec struct {
...
+	// Contains all the configuration that applies to both host and member operators
+	// +optional
+	Global GlobalConfig `json:"global,omitempty"`
+}
+
+// GlobalConfig contains all configuration that applies for both host and member operators
+// +k8s:openapi-gen=true
+type GlobalConfig struct {
+	// +optional
+	PublicViewer PublicViewerConfig `json:"publicViewer,omitempty"`
```
mfrancisc, line 45: It seems like a good idea to start introducing a Global/Common config but I'm wondering if we could add the PublicViewer struct to both HostConfig and Members, instead of creating a Global config? Is there any benefit for having this?

api/v1alpha1/usersignup_types.go
```diff
@@ -92,6 +92,7 @@ const (
 	UserSignupPendingApprovalReason            = "PendingApproval"
 	UserSignupUserBanningReason                = "Banning"
 	UserSignupUserBannedReason                 = "Banned"
+	UserSignupUserReservedReason               = "Reserved"
```
mfrancisc, line 95: I think this is for checking that the "public-viewer" MUR is not reused by some "real" user right? Maybe this reason is more suitable in the masteruserrecord_types ? WDYT?

mfrancisc (commented):
Nice 👍 , thanks for creating the PRs!

I have few comments.

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -32,6 +32,30 @@ type ToolchainConfigSpec struct {
...
+
+// PublicViewer used to enable public-viewer support
+// +k8s:openapi-gen=true
+type PublicViewerConfig struct {
+	// +kubebuilder:default:=false
+	Enabled bool `json:"enabled"`
+
+	// +kubebuilder:default:=public-viewer
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
```
MatousJobanek, line 56: max should be 63 - it's used as a label value of the SpaceBindings 
```suggestion
	// +kubebuilder:validation:MaxLength=63
```

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -32,6 +32,30 @@ type ToolchainConfigSpec struct {
...
+	// +kubebuilder:validation:Pattern=`^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
```
MatousJobanek, line 57: out of curiosity - where did you get the regex from - looks a bit overcomplicated. Could be maybe simplified to:
```
^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
```
https://github.com/kubernetes/kubernetes/blob/20d0ab7ae808aaddb1556c3c38ca0607663c50ac/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L177-L183

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -32,6 +32,30 @@ type ToolchainConfigSpec struct {
...
+	PublicViewer PublicViewerConfig `json:"publicViewer,omitempty"`
+}
+
+// PublicViewer used to enable public-viewer support
+// +k8s:openapi-gen=true
+type PublicViewerConfig struct {
+	// +kubebuilder:default:=false
+	Enabled bool `json:"enabled"`
+
+	// +kubebuilder:default:=public-viewer
```
MatousJobanek, line 54: default should be empty, right?

api/v1alpha1/toolchainconfig_types.go
filariow: I got that from the [gateway-api pkg](https://github.com/kubernetes-sigs/gateway-api/blob/817292143c4f59a88c955c59c8c1712560d67a79/apis/v1alpha2/shared_types.go#L157), but it seems that the simpler one you are sharing suites way better.

Btw, is it something we'd like to have to avoid IP addresses as names here?

For sure, we don't want `*.something` as a valid name, I didn't notice that 😅

api/v1alpha1/toolchainconfig_types.go
filariow: This configuration is used by host operator and registration service. Neither HostConfig nor MembersConfig seemed appropriate to me.
Adding the configuration in both of them does not sound a good idea to me. It would become possible to have two different values for the same configuration and cause misbehavior.

As @MatousJobanek suggested [here](https://github.com/codeready-toolchain/api/pull/410#discussion_r1537253437) we could add this config under `HostConfig`. My only concern about going this path is related to Registration Service's future improvements. Indeed, as of now, the RegistrationService is thought to be installed on the Host cluster only, but it seems to me that it may be easily decoupled so that it can be executed on member clusters too. So, adding this configuration under `HostConfig` may cause breaking changes or duplication if we decide to decouple the RegistrationService from the Host Cluster. WDYT?

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -32,6 +32,30 @@ type ToolchainConfigSpec struct {
 	// Contains all member operator configurations for all member clusters
 	// +optional
 	Members Members `json:"members,omitempty"`
+
+	// Contains all the configuration that applies to both host and member operators
+	// +optional
+	Global GlobalConfig `json:"global,omitempty"`
```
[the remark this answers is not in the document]
filariow: continuing discussion [here](https://github.com/codeready-toolchain/api/pull/410#discussion_r1537377816)

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -32,6 +32,30 @@ type ToolchainConfigSpec struct {
...
+type GlobalConfig struct {
+	// +optional
+	PublicViewer PublicViewerConfig `json:"publicViewer,omitempty"`
+}
+
+// PublicViewer used to enable public-viewer support
+// +k8s:openapi-gen=true
+type PublicViewerConfig struct {
+	// +kubebuilder:default:=false
+	Enabled bool `json:"enabled"`
```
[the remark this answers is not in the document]
filariow: the username would be sufficient, but not explicit. Hence, the purpose of the boolean `enabled` field is to make the feature flag explicit

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -32,6 +32,30 @@ type ToolchainConfigSpec struct {
...
+
+	// +kubebuilder:default:=public-viewer
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	Username string `json:"username,omitempty"`
```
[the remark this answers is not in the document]
filariow: fixed in 7225626b4936c8f72f50eaad24bc272bee268344

api/v1alpha1/toolchainconfig_types.go
filariow: resolved in 7225626b4936c8f72f50eaad24bc272bee268344

api/v1alpha1/toolchainconfig_types.go
filariow: resolved in 7225626b4936c8f72f50eaad24bc272bee268344

api/v1alpha1/toolchainconfig_types.go
filariow: resolved in 7225626b4936c8f72f50eaad24bc272bee268344

api/v1alpha1/spacebindingrequest_types.go
```diff
@@ -26,6 +26,9 @@ const (
 
 	// SpaceBindingRequestProvisionedReason represents the reason for a successfully provisioned space binding request.
 	SpaceBindingRequestProvisionedReason = provisionedReason
+
+	// SpaceBindingRequestInvalidReason represents the reason for an invalid space binding request.
+	SpaceBindingRequestInvalidReason = invalidReason
```
rajivnathan, line 31: Shouldn't the creation of the resource be blocked if it's invalid? Why allow creation of an invalid resource?

filariow:
@rajivnathan sorry, I forgot to add the link. Added it in the description.

api/v1alpha1/toolchainconfig_types.go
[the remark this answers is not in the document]
sadlerap: An alternative we could pursue down the line would be to have `GlobalConfig` have a pointer to an instance of this struct, with the nil case indicating that this particular feature isn't enabled.  However, this has the downside of being less explicit about which features are enabled, and it breaks the encapsulation of `PublicViewerConfig` (i.e. if more stuff needs to be added to this struct orthogonal to the user, then we're going to need to reconsider our approach).

api/v1alpha1/spacebindingrequest_types.go
filariow: great point, this could be implemented in the [ValidatingWebhook on member cluster](https://github.com/codeready-toolchain/member-operator/blob/1f0cd92ffd612099ddf238da412010c1e3f1e3ec/pkg/webhook/validatingwebhook/validate_spacebindingrequest.go#L28). WDYT?

api/v1alpha1/usersignup_types.go
filariow: I think this is just something I forgot to remove, as it's not used. That was used before figuring out that `forbiddenUsername` could help here. let me remove this.

filariow: removed in ac271dff1ebc1de3130e6ed8335b9d6db9bdc999

api/v1alpha1/toolchainconfig_types.go
alexeykazakov: > the RegistrationService is thought to be installed on the Host cluster only, but it seems to me that it may be easily decoupled so that it can be executed on member clusters too

Why would you want to install the reg service on a member cluster? I don't think we ever considered that. But maybe I missed something?
What would make sense is to decouple the proxy and reg-service so the proxy can be an optional component but it's still all about the host cluster components. Isn't it?

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -32,6 +32,30 @@ type ToolchainConfigSpec struct {
...
+// +k8s:openapi-gen=true
+type PublicViewerConfig struct {
+	// +kubebuilder:default:=false
+	Enabled bool `json:"enabled"`
+
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	Username string `json:"username,omitempty"`
```
alexeykazakov, line 58: What is this username used for? Can you please add a comment here describing this field.

alexeykazakov:
> I forgot to add the link. Added it in the description.

Thanks for the JIRA link. However the JIRA and the attached feature refinement doc describes the high level desired UX. What would be really useful is a detailed technical design doc. It would help a lot with the review of the PR.
This is how we usually handle development of new features (unless it's a very simple feature):
1. Some description and agreement of the overall desired UX. In this case it's done with the feature refinement doc.
2. A technical design doc where we can discuss and agree on the overall approach of how we implement the feature. This is what is missing here.
3. PRs with an implementation of that design.

api/v1alpha1/toolchainconfig_types.go
filariow: Sorry, probably I didn't express myself correctly. I'm not saying that's something we are planning or we should plan. I'm saying that the fact that it's technically doable, makes me feel like we can find better place where to put this configuration 😄

filariow:
superseded by #430

PR #414, opened 2024-04-21
Adding domains field to auto-approval config

dperaza4dustbit:
Implements: [ASC-444](https://issues.redhat.com//browse/ASC-444)

## Description
Adding a domain list to the auto-approval section in toolchain config resource. List will be used to allow
auto-approval for emails under specific domains. 

## Checks
1. Did you run `make generate` target? yes

2. Did `make generate` change anything in other projects (host-operator, member-operator)? yes

3. In case of **new** CRD, did you the following? NA

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/1013
    - member-operator: NA


api/v1alpha1/toolchainconfig_types.go
```diff
@@ -102,6 +102,11 @@ type AutomaticApprovalConfig struct {
 	// Defines if the automatic approval is enabled or not
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// Email domains to consider for auto-approval.
+	// +optional
+	// +listType=set
+	Domains []string `json:"domains,omitempty"`
```
MatousJobanek, line 109: We already have some parameters for specifying a set of domains, in this case a string where the list of domains are comma-separated values - see: https://github.com/codeready-toolchain/api/blob/d9f5a3d7a654c4ea17e96dd127ce35f3cfd059c4/api/v1alpha1/toolchainconfig_types.go#L144-L148
the parameter is then read in this way:
https://github.com/codeready-toolchain/host-operator/blob/6a7e6922bd6f74fe0e35b647e9b8908a716f1254/controllers/toolchainconfig/configuration.go#L183-L189
How about using the same approach/format?

dperaza4dustbit: Sure, that can work too, do you recall the advantage of using comma separated string instead of a set?

dperaza4dustbit: something else I would avoid is call the field domains and the expect something like this @redhat.com when the domain is really redhat.com. Could cause confusion.

dperaza4dustbit: OK, lets keep it consistent like you suggest. Making the change

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -102,6 +102,11 @@ type AutomaticApprovalConfig struct {
 	// Defines if the automatic approval is enabled or not
 	// +optional
 	Enabled *bool `json:"enabled,omitempty"`
+
+	// Comma-separated email domains to consider for auto-approval.
+	// For example: "redhat.com,ibm.com"
```
alexeykazakov, line 107: What's the difference between the following cases:
```
automaticApprovalConfig:
    enabled: true
    domains: redhat.com,ibm.com
```
vs
```
automaticApprovalConfig:
    enabled: false
    domains: redhat.com,ibm.com
```
vs.
```
automaticApprovalConfig:
    enabled: true
```

dperaza4dustbit: good question, I was thinking this:
enabled: true --> Allows for auto-approval of all authenticated emails 

enable: true and domains: redhat.com,ibm.com --> Allows for auto-approval of all authenticated users with emails under those domains

enable: false and domains: redhat.com,ibm.com --> does not auto-approve and ignores domain

alexeykazakov: So basically the default value for the domains is "*". And the "enabled" true|false dictates if the approval (for "*" or specific domains) is enabled or not. I think it's reasonable approach. Could you please add it as a description to the source code here?

dperaza4dustbit: OK, I can add more description here on API changes, planning to add more details when the actual logic is implemented on host side.

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -97,6 +97,14 @@ type AutomaticApprovalConfig struct {
...
```
alexeykazakov, line 102:
```suggestion
	// For example: "domain.com,anotherdomain.org"
```

alexeykazakov (approved):
Looks good. Thanks! Not sure about the example domains though :)

api/v1alpha1/toolchainconfig_types.go
dperaza4dustbit: Will make this change thanks for suggestion, yeah lets keep me more generic

MatousJobanek (approved):

PR #409, opened 2024-03-18
Update K8s dependencies to v0.27

ranakan19:
## Description
DO-NOT Merge
(All the PRs related to version updates need to be merged at once, but these PR are for early feedback so that there are not a lot of changes to review at once)

Updates:
- Kubernetes dependencies to v0.27
- Controller-gen to v0.12
- Controller runtime to v0.15

## Checks
1. Did you run `make generate` target? **yes**

2. Did `make generate` change anything in other projects (host-operator, member-operator)? **yes**

3. In case of **new** CRD, did you the following? **n/a**
    - make/generate.mk in this repository
    - `resources/setup/roles/host.yaml` in the sandbox-sre repository
    - `PROJECT` file: https://github.com/codeready-toolchain/host-operator/blob/master/PROJECT
    - `CSV` file: https://github.com/codeready-toolchain/host-operator/blob/master/config/manifests/bases/host-operator.clusterserviceversion.yaml

4. In case other projects are changed, please provides PR links. 
https://github.com/codeready-toolchain/host-operator/pull/994
https://github.com/codeready-toolchain/member-operator/pull/546

PR #421, opened 2024-04-30
bump kube-openapi version

filariow:
## Description
Update kube-openapi version and removes Fake GOPATH implementation

## Checks

4. In case other projects are changed, please provides PR links. **NA**


mfrancisc (approved):
Thanks for updating the openapi tool 🙏 
I've tested and it works fine for me. 

My folder structure is pretty standard so let's wait for others to test it as well before merging.

rajivnathan (approved):
Nice, worked well for me too

fbm3307 (approved):
Worked for me

go.mod
```diff
@@ -10,48 +10,48 @@ require (
 	k8s.io/apimachinery v0.25.0
 	k8s.io/code-generator v0.25.0
 	k8s.io/gengo v0.0.0-20211129171323-c02415ce4185
-	k8s.io/kube-openapi v0.0.0-20220803162953-67bda5d908f1
+	k8s.io/kube-openapi v0.0.0-20240430033511-f0e62f92d13f
```
fbm3307, line 13: Just a Thought , Do we need to discuss/resolve the update to kube-openapi version as i see PR https://github.com/codeready-toolchain/api/pull/409/files from Kanika  updating it to some different version , and here it updates to some latest version? wont that be conflicting or may be just duplication of work?

filariow: Great catch! As kube-openapi is a [controller-runtime indirect dependency](https://github.com/kubernetes-sigs/controller-runtime/blob/9bc967a711e9538cff4099645dd6df64c9ba8a74/go.mod#L94), maybe it's worth to merge the other PR first and then just remove the FAKE GOPATH logic with this one.

ranakan19: Thanks @fbm3307! I was thinking the same, but I wonder why there are no errors with the kube-openapi version you are upgrading to (which is the latest version) with the older controller-runtime version. Looking into this further. 
As for merging the PR - I cannot merge the other PR untill all the other PRs across all repos are ready - so it may take another sprint or two to get to that stage. Would it be okay to wait?

ranakan19: Ah i see, the version you are updating to is the latest one from controller-runtime, not from the version of the controller-runtime we are currently on or would like to update it. It also has a higher version golang version. Since there is no error in upgrading to the latest version, because it is an indirect dependency, you could update the kube-openapi to this version if you need it for something. If not, I'd rather keep them updated in sync with the rest of the dependencies. 
The version of kube-openapi for controller-runtime v0.15 is https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.15/go.mod#L71

filariow: I agree, we should definitely keep them updated in sync. Unfortunately, we'll not being able to remove FAKE GOPATH implementation until -if I'm not mistaken- controller-runtime@v0.18.

fbm3307:
@filariow probably we should put the PR to draft and pick up later or close this ( as per the above discussion)

PR #425, opened 2024-05-27
update webhook deploy comment

mfrancisc:
## Description
Update the `WebhookConfig.Deploy` field which now is used in a different way. Based on the changes [in this PR](https://github.com/codeready-toolchain/member-operator/pull/562) and related ones, the webhook will still be deployed but deleted afterwords.

Those PRs will be merged only after the new behavior is merged to master. 

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/1036
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/572

### DIFF
--- a/api/v1alpha1/memberoperatorconfig_types.go
+++ b/api/v1alpha1/memberoperatorconfig_types.go
@@ -156,7 +156,7 @@ type ToolchainClusterConfig struct {
 // +k8s:openapi-gen=true
 type WebhookConfig struct {
 	// Defines the flag that determines whether to deploy the Webhook.
-	// If the deploy flag is disabled, the Webhook is deployed and deleted immediately after by the memberoperatorconfig controller.
+	// If the deploy flag is set to False and the Webhook was deployed previously it will be deleted by the memberoperatorconfig controller.
 	// +optional
 	Deploy *bool `json:"deploy,omitempty"`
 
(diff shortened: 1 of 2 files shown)


api/v1alpha1/memberoperatorconfig_types.go
```diff
@@ -155,7 +155,8 @@ type ToolchainClusterConfig struct {
 // Defines all parameters concerned with the Webhook
 // +k8s:openapi-gen=true
 type WebhookConfig struct {
-	// Defines the flag that determines whether to deploy the Webhook
+	// Defines the flag that determines whether to deploy the Webhook.
+	// If the deploy flag is disabled, the Webhook is deployed and deleted immediately after by the memberoperatorconfig controller.
```
rajivnathan, line 159: It kind of sounds like we're saying the webhook will be deployed and then deleted again. How about rewording it like this to make it a bit more clear?

```suggestion
	// If the deploy flag is set to False and the Webhook was deployed previously it will be deleted by the memberoperatorconfig controller.
```

rajivnathan (approved):
Thanks for updating the comment 👍 Just made a small suggestion regarding the wording

api/v1alpha1/memberoperatorconfig_types.go
mfrancisc: yup, makes more sense indeed!

rajivnathan (approved):

PR #429, opened 2024-06-18
API for feature toggles

alexeykazakov:
## Description

API for feature toggles/flags.

https://issues.redhat.com/browse/SANDBOX-653

## Checks

3. In case of **new** CRD, did you do the following? **yes**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/1044
    - member-operator: N/A


api/v1alpha1/toolchainconfig_types.go
```diff
@@ -469,6 +477,36 @@ type TiersConfig struct {
...
+	// For example if there are two features:
+	// - feature1, weight=5
+	// - feature2, weight=90
+	// And tiers (one or many) contain the following object manifests:
+	// - RoleBinding with "toolchain.dev.openshift.com/feature: feature1" annotation
+	// - ConfigMap with "toolchain.dev.openshift.com/feature: feature2" annotation
+	// Then the RoleBinding will be created for the corresponding tiers with probability of 0.05 (around 5 out of every 100 spaces would have it)
+	// And the ConfigMap will be created with probability of 0.9 (around 90 out of every 100 spaces would have it)
+	// +optional
+	Weight *int `json:"weight,omitempty"`
```
mfrancisc, line 507: it looks like it's a percentage value, another option would be to name this `WeightPercentage`

mfrancisc (approved):
Looks great! the comments are very clear and well written 👍

api/v1alpha1/toolchainconfig_types.go
MatousJobanek: yup, agree with @mfrancisc 
let's also add kubebuilder validation markers plus you can use `uint` (instead of `int`) to exactly match the usage 
```suggestion
	// +kubebuilder:validation:Minimum=0
	// +kubebuilder:validation:Maximum=100
	WeightPercentage *uint `json:"weightPercentage,omitempty"`
```

alexeykazakov: I didn't want to use percentages because it's not really %
I'm concerned about misleading here. We do not guarantee that % of users will get the feature. We don't take into account how many users already have this feature for example. It's rather a probability of winning the "lottery" multiplied by 100. This is why I just call it weight which is genetic enough.

alexeykazakov: But good point about the validation and the uint :+1:

MatousJobanek: yeah, good point, I misread the description, sorry about the confusion and thanks for the clarification.

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -469,6 +480,37 @@ type TiersConfig struct {
...
+	// - feature1, weight=5
+	// - feature2, weight=90
+	// And tiers (one or many) contain the following object manifests:
+	// - RoleBinding with "toolchain.dev.openshift.com/feature: feature1" annotation
+	// - ConfigMap with "toolchain.dev.openshift.com/feature: feature2" annotation
+	// Then the RoleBinding will be created for the corresponding tiers with probability of 0.05 (around 5 out of every 100 spaces would have it)
+	// And the ConfigMap will be created with probability of 0.9 (around 90 out of every 100 spaces would have it)
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=100
```
MatousJobanek, line 510: You can also add a marker for the default value:
```suggestion
	// +kubebuilder:validation:Maximum=100
	// +kubebuilder:default=100
```

MatousJobanek (approved):
Looks great :+1:

api/v1alpha1/toolchainconfig_types.go
MatousJobanek: tbh, we should do this for all fields in the ToolchainConfig, so it's automatically included in the generated godoc and we can use it as a reference. But that's for later

api/v1alpha1/toolchainconfig_types.go
mfrancisc: Ok, I see, it's a _probability_ of getting the feature, not a percentage, I misunderstood the comment. thanks for the clarification 👍

ranakan19 (approved):
Looks great

PR #433, opened 2024-07-04
Move info into tc status

metlos:
## Description
This introduces new fields into the ToolchainCluster status: `apiEndpoint` and `operatorNamespace` are introduced into the status because these fields will be read from the kubeconfig contained in the ToolchainCluster secret instead of explicitly being set in the ToolchainCluster's spec.

## Checks

3. In case of **new** CRD, did you the following? **no new CRD**

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/1056
    - member-operator: https://github.com/codeready-toolchain/member-operator/pull/583


api/v1alpha1/toolchaincluster_types.go
```diff
@@ -66,6 +79,13 @@ type LocalSecretReference struct {
 // cluster updated periodically by cluster controller.
 // +k8s:openapi-gen=true
 type ToolchainClusterStatus struct {
+	// APIEndpoint is the API endpoint of the remote cluster. This can be a hostname,
+	// hostname:port, IP or IP:port.
+	APIEndpoint string `json:"apiEndpoint,omitempty"`
+
+	// OperatorNamespace is the namespace in which the operator runs in the remote cluster
+	OperatorNamespace string `json:"operatorNamespace,omitempty"`
```
mfrancisc, line 87: minor - but usually we set `omitempty` with optional fields only.
```suggestion
	// APIEndpoint is the API endpoint of the remote cluster. This can be a hostname,
	// hostname:port, IP or IP:port.
	APIEndpoint string `json:"apiEndpoint"`

	// OperatorNamespace is the namespace in which the operator runs in the remote cluster
	OperatorNamespace string `json:"operatorNamespace"`
```

mfrancisc (commented):
Looks good 👍 

I have only one small comment.

mfrancisc (approved):
Thanks for the additional notes and comments on the fields 👍

api/v1alpha1/toolchaincluster_types.go
```diff
@@ -66,6 +79,13 @@ type LocalSecretReference struct {
 // cluster updated periodically by cluster controller.
 // +k8s:openapi-gen=true
 type ToolchainClusterStatus struct {
+	// APIEndpoint is the API endpoint of the remote cluster. This can be a hostname,
+	// hostname:port, IP or IP:port.
+	APIEndpoint string `json:"apiEndpoint"`
+
+	// OperatorNamespace is the namespace in which the operator runs in the remote cluster
+	OperatorNamespace string `json:"operatorNamespace"`
```
MatousJobanek, line 87: You need to make these fields optional - there are three reasons:
1. if you kept it as it is, then all already existing ToolchainCluster CRs would become invalid (as they don't have the fields set). I'm not user if OLM would be able to update the CRD at all.
2. ToolchainCluster controller doesn't have the logic that populates these fields yet, so it would not be able to update the ToolchainClusterStatus at all (rejected as the fields are required) until the logic is added.
3. In the future, if the ToolchainCluster is provided with either non-existing or invalid secret/kubeconfig, then we would need to set the "ready" condition to false and leave these two fields unset (empty)
```suggestion
	// hostname:port, IP or IP:port.
	// +optional
	APIEndpoint string `json:"apiEndpoint,omitempty"`

	// OperatorNamespace is the namespace in which the operator runs in the remote cluster
	// +optional
	OperatorNamespace string `json:"operatorNamespace,omitempty"`
```

MatousJobanek: Just looking at the e2e tests in the other PRs and yes, they both failed in the migration test suite:
```
  - lastTransitionTime: "2024-07-04T19:47:13Z"
    message: 'error validating existing CRs against new CRD''s schema for "toolchainclusters.toolchain.dev.openshift.com":
      error validating toolchain.dev.openshift.com/v1alpha1, Kind=ToolchainCluster
      "toolchain-host-04194050/member-ci-op-80dsm6yr-1de72.devsandboxci.devcluster.openshift.1":
      updated validation is too restrictive: [[].status.apiEndpoint: Required value,
      [].status.operatorNamespace: Required value]'
    reason: InstallComponentFailed
    status: "True"
    type: InstallPlanFailed
```

mfrancisc: Ohhh jeez, I haven't thought about this yesterday when reviewing the PR - that we cannot introduce required fields into existing CRs 🤦‍♂️ . Great point @MatousJobanek !

metlos: Yeah, I was suspecting that but wasn't 100% how k8s would behave, so I tried. CI seems to be failing with OLM not being able to upgrade.

api/v1alpha1/toolchaincluster_types.go
MatousJobanek, line 87: Is there any reason why you didn't include the `// +optional` marker?
```suggestion
	// hostname:port, IP or IP:port.
	// +optional
	APIEndpoint string `json:"apiEndpoint,omitempty"`

	// OperatorNamespace is the namespace in which the operator runs in the remote cluster
	// +optional
	OperatorNamespace string `json:"operatorNamespace,omitempty"`
```

MatousJobanek (approved):

api/v1alpha1/toolchaincluster_types.go
```diff
@@ -66,6 +79,15 @@ type LocalSecretReference struct {
...
 // +k8s:openapi-gen=true
 type ToolchainClusterStatus struct {
+	// APIEndpoint is the API endpoint of the remote cluster. This can be a hostname,
+	// hostname:port, IP or IP:port.
+	// +optional
+	APIEndpoint string `json:"apiEndpoint"`
+
+	// OperatorNamespace is the namespace in which the operator runs in the remote cluster
+	// +optional
+	OperatorNamespace string `json:"operatorNamespace"`
```
MatousJobanek, line 89: in other words, this will ensure that the fields are optional, but they will stay visible (empty) in the status part if they are not set. Yup, that's fine. In most cases, we include the `omitempty` together with `+optional`, but it's not critical in this case.

api/v1alpha1/toolchaincluster_types.go
```diff
@@ -34,20 +34,33 @@ const (
 type ToolchainClusterSpec struct {
 	// The API endpoint of the member cluster. This can be a hostname,
 	// hostname:port, IP or IP:port.
+	//
+	// Be aware that this field is going to be replaced with
+	// the Status.APIEndpoint in the future.
 	APIEndpoint string `json:"apiEndpoint"`
```
MatousJobanek, line 40: btw, you could have made this field optional as well, so you can drop populating it before you remove it from the API

MatousJobanek (approved):

api/v1alpha1/toolchaincluster_types.go
metlos: IMHO, these two fields are a vital part of the information that ToolchainCluster provides, so I wanted them to be prominent, especially in edge cases where they would not be set.

MatousJobanek: :+1:

PR #437, opened 2024-07-24
KUBESAW-129: Makefile target to check compatibility of local version with other repos

fbm3307:
## Description
This second part of dependency check

There is a makefile target that validates compatibility of the local version in other repositories that rely on them

Similar-Related PR in Toolchain-common - https://github.com/codeready-toolchain/toolchain-common/pull/417

## Checks
1. Did you run `make generate` target? **NA**

2. Did `make generate` change anything in other projects (host-operator, member-operator)? **NA**

    NA


make/go.mk
```diff
@@ -13,3 +13,31 @@ GO_PACKAGE_PATH ?= github.com/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}
...
+	$(eval REPO_PATH = ${BASE_REPO_PATH}/$(shell basename $(repo))) \
+	git clone --depth=1 $(repo) ${REPO_PATH}; \
+	cd ${REPO_PATH}; \
+	if [[ "${GH_HOST}" == "$(repo)" ]]; then \
+		$(MAKE) generate ; \
+	elif [[ "${GH_MEMBER}" == "$(repo)" ]]; then\
+		$(MAKE) generate-assets ; \
+	fi; \
+	go mod edit -replace github.com/codeready-toolchain/api=${C_PATH}; \
+	$(MAKE) verify-dependencies; \
```
MatousJobanek, line 41: correct me if I'm wrong, but if some of the verification fails, then the for loop ends as well and it won't validate the other repositories. 
We need to make sure that:
* all repositories are validated
* the target fails if at least one validation fails

make/go.mk
```diff
@@ -13,3 +13,31 @@ GO_PACKAGE_PATH ?= github.com/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}
...
+GH_REGSVC = $(GH_BASE_URL_CRT)registration-service
+GH_E2E = $(GH_BASE_URL_CRT)toolchain-e2e
+GH_TC = $(GH_BASE_URL_CRT)toolchain-common
+
+.PHONY: verify-replace-run
+verify-replace-run: generate
+	$(eval C_PATH = $(PWD))\
+	$(foreach repo,${GH_HOST} ${GH_MEMBER} ${GH_REGSVC} ${GH_E2E} ${GH_TC} ${GH_KSCTL},\
+	$(eval REPO_PATH = ${BASE_REPO_PATH}/$(shell basename $(repo))) \
+	git clone --depth=1 $(repo) ${REPO_PATH}; \
```
MatousJobanek, line 33: this is a mix of makefile operands with embedded bash script which is a bit hard to read. It would be better to have everything in the bash format.
feel free to move the logic to the bash completely and only invoke the script from the target

make/go.mk
fbm3307: No this does not end the for loop, it shows the error and moves to other repo, it  validates other repos and even if one validation fails the target fails

make/go.mk
```diff
@@ -13,3 +13,31 @@ GO_PACKAGE_PATH ?= github.com/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}
 build:
 	$(Q)CGO_ENABLED=0 GOARCH=amd64 GOOS=linux \
 	    $(GO) build github.com/codeready-toolchain/api/api/v1alpha1/
+
+TMP_DIR = /tmp/
+BASE_REPO_PATH = $(shell mktemp -d ${TMP_DIR}crt-verify.XXX)
```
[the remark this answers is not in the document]
fbm3307: This wont give the path to /tmp folder right ? we want to have these mktemp in /tmp .

make/go.mk
```diff
@@ -13,3 +13,31 @@ GO_PACKAGE_PATH ?= github.com/${GO_PACKAGE_ORG_NAME}/${GO_PACKAGE_REPO_NAME}
...
+	$(eval C_PATH = $(PWD))\
+	$(foreach repo,${GH_HOST} ${GH_MEMBER} ${GH_REGSVC} ${GH_E2E} ${GH_TC} ${GH_KSCTL},\
+	$(eval REPO_PATH = ${BASE_REPO_PATH}/$(shell basename $(repo))) \
+	git clone --depth=1 $(repo) ${REPO_PATH}; \
+	cd ${REPO_PATH}; \
+	if [[ "${GH_HOST}" == "$(repo)" ]]; then \
+		$(MAKE) generate ; \
+	elif [[ "${GH_MEMBER}" == "$(repo)" ]]; then\
+		$(MAKE) generate-assets ; \
+	fi; \
```
MatousJobanek, line 39: I thought that we wanted to avoid this if-else statement by executing it as part of the `verify-dependencies` target

make/go.mk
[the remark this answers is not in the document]
MatousJobanek: it does:
```
$ mktemp -d --suffix=-replace-verify
/tmp/tmp.xlrn7wIPwR-replace-verify
```

make/go.mk
MatousJobanek: > No this does not end the for loop, it shows the error and moves to other repo, it validates other repos

ok, that's cool, I didn't know about that. Thanks for clarifying 

> and even if one validation fails the target fails

I just tried that and the returned value of the target is `0` even when some of the validation failed

MatousJobanek: apart from that, it would be nice to list the repositories the validation failed for (and ideally also the actual errors); otherwise, it can be a bit challenging going through the whole output.
You can try to play with the stderr output of the validations and/or with the returned status values

make/go.mk
fbm3307: yes.. the main thing is we do the `verify-dependencies` after `go mod replace` , but this  `generate` is required before the `go mod replace`..  for host(for tiers) and member(for webhooks) explicitly ,
so one thing i can do is have one another makefile target command  `prerequisite`  in all the repos,    apart from host and member, in all the repos this command will be empty. 
Due to this reason i wrote if else but i can change it now WDYT..?

MatousJobanek: really? it doesn't seem so.
What I see is that this is needed to run `go mod tidy`. Running `go mod edit -replace` in a cloned repository works without any problem. It fails for `go mod tidy` (as part of `verify-dependencies`) which means that the `verify-dependencies` target in host & member repos should be fixed - we need to run `generate` targets before `go mod tidy`

fbm3307: more discussion of this on slack https://redhat-internal.slack.com/archives/C06MJ2DVBU4/p1721916497251399

make/go.mk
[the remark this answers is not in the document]
fbm3307: I tried, this but mac os has a lower linus version thank the fedora. so the suffix itself is not available 
it requires a path to be given

make/go.mk
fbm3307: Done

mfrancisc (approved):
Great job 👏

MatousJobanek (approved):

MatousJobanek (approved):
Great job :+1: Thanks for addressing all the comments and moving it to a script. This is much easier to read and maintain :1st_place_medal:

PR #440, opened 2024-08-15
Add target cluster to social event

alexeykazakov:
Replaces `PreferSameCluster`, which is not currently used at all, by `TargetCluster`

Part of https://issues.redhat.com/browse/SANDBOX-724

- host-operator: https://github.com/codeready-toolchain/host-operator/pull/1075


ranakan19 (approved):

rajivnathan (approved):

mfrancisc (commented):
Just curious - I guess we are safe with changing the name here since the existing socialevents CRs do not use this field , thus no migration code is needed here .

alexeykazakov:
> Just curious - I guess we are safe with changing the name here since the existing socialevents CRs do not use this field , thus no migration code is needed here .

Correct. The samecluster field was not used anywhere. Controllers do not use it (so it was ignored). Existing social event CRs didn't have this field set either. So no migration needed.

PR #441, opened 2024-08-29
feat: Add new field for TemplateObjects in TierTemplate CRD to support go template syntax

mfrancisc:
## Description
This PR introduces a new field which will enable us later to use go template syntax later for defining the tier templates.
It is the first step in the [Tier Decoupling effort](https://docs.google.com/document/d/1x5SoBT80df9fmVsaDgAE6DE7hE6lzmNIK087JUmgaJs/edit)

Jira: https://issues.redhat.com/browse/KUBESAW-135

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/1080


api/v1alpha1/tiertemplate_types.go
```diff
@@ -28,7 +29,19 @@ type TierTemplateSpec struct {
 	Revision string `json:"revision"`
 
 	// Template contains an OpenShift Template to be used to provision either a user's namespace or cluster-wide resources
+	// Note: this field will be removed in favor of the new TemplateObjects below.
```
alexeykazakov, line 32: Then make it's optional? You you want to make it optional later?

api/v1alpha1/tiertemplate_types.go
```diff
@@ -28,7 +29,19 @@ type TierTemplateSpec struct {
...
+
+	// TemplateObjects contains list of Unstructured Objects that can be parsed at runtime and will be applied as part of the tier provisioning.
+	//
+	// NOTE: when specifying variables as part of the objects list , those concatenated as part of other strings do not need to be wrapped inside quotes,
+	// while those that are not part of other strings do need to be wrapped in single quotes. This is required otherwise the yaml parser will error while trying to parse those resources containing variables.
+	// eg: https://docs.google.com/document/d/1x5SoBT80df9fmVsaDgAE6DE7hE6lzmNIK087JUmgaJs/edit#heading=h.2iuytpfnmul5
+	//
+	// The template parameters values will be defined in the NSTemplateTier CRD.
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
```
alexeykazakov, line 43: I think we need to define the list type.
It should be:
```
// +listType=set
```
or 
```
// +listType=atomic
```
We can't use `map` here.

api/v1alpha1/tiertemplate_types.go
```diff
@@ -28,7 +29,19 @@ type TierTemplateSpec struct {
...
+	TemplateObjects []runtime.RawExtension `json:"templateObjects" protobuf:"bytes,3,rep,name=templateObjects"`
```
alexeykazakov, line 44:
```suggestion
	TemplateObjects []runtime.RawExtension `json:"templateObjects,omitempty" protobuf:"bytes,3,rep,name=templateObjects"`
```
?

api/v1alpha1/tiertemplate_types.go
fbm3307, line 43: Adding `// +listType=set` would be a better option here

api/v1alpha1/tiertemplate_types.go
fbm3307, line 44:
```suggestion
	TemplateObjects []runtime.RawExtension `json:"templateObjects" protobuf:"bytes,3,opt,name=templateObjects"`
```
Can we use opt here instead of rep?

fbm3307: +1 to adding `omitempty`

api/v1alpha1/tiertemplate_types.go
mfrancisc: yes, my plan was to make it optional later once we get ready for removing it.

api/v1alpha1/tiertemplate_types.go
mfrancisc: why can't we use map ? 

Also I don't see the listType defined in the openshift templates type https://github.com/openshift/api/blob/5852b58f4b1071fe85d9c49dff2667a9b2a74841/template/v1/types.go#L39-L40 , but maybe they are doing something slightly different there.

api/v1alpha1/tiertemplate_types.go
mfrancisc: thanks

api/v1alpha1/tiertemplate_types.go
mfrancisc: I'm leaning towards `set` :
```
//	+listType=set
//
// Sets are lists that must not have multiple times the same value. Each
// value must be a scalar (or another atomic type).
//
```
but not sure if we might want to allow duplicated objects in the templates, or if it makes sense. The `atomic` type seems less performant .

alexeykazakov: > but not sure if we might want to allow duplicated objects in the templates

I might be missing something but I can't imagine a situation when we would need to keep 100% identical duplicated objects in the templates.

api/v1alpha1/tiertemplate_types.go
fbm3307: `rep` can be used if we think that the objects would be repeated , i this case since its a list of Template Objects, i think rep would be fine

mfrancisc: I've accepted your suggestion, thanks.

api/v1alpha1/tiertemplate_types.go
mfrancisc: done in https://github.com/codeready-toolchain/api/pull/441/commits/fbf4de1ac839507f95eb88fdb11b827f6aa7bbe3

mfrancisc: added the list type in https://github.com/codeready-toolchain/api/pull/441/commits/fbf4de1ac839507f95eb88fdb11b827f6aa7bbe3

fbm3307 (approved):
looks good
Thank you for addressing my comments 👍

alexeykazakov (approved):

api/v1alpha1/tiertemplate_types.go
mfrancisc: ok, the `atomic` type worked.

PR #446, opened 2024-09-23
feat: Add new TierTemplateRevision CRD

mfrancisc:
## Description
Adds a new CRD named TierTemplateRevision CRD, which is will be used to store the revisions of the various TierTemplates when something changes.

Jira: https://issues.redhat.com/browse/KUBESAW-138

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/#


make/generate.mk
```diff
@@ -5,7 +5,7 @@ API_VERSION:=v1alpha1
 
 # how to dispatch the CRD files per repository (space-separated lists)
 # !!! IMPORTANT !!! - when there is a new CRD added or an existing one removed or renamed, don't forget to change it also here: https://github.com/codeready-toolchain/toolchain-cicd/blob/master/scripts/add-cluster.sh#L52-L73
```
alexeykazakov, line 7: This is not relevant anymore, right? Let's delete this comment.

alexeykazakov (approved):

make/generate.mk
mfrancisc: Good catch! Removed in https://github.com/codeready-toolchain/api/pull/446/commits/27efbca33495309e7d214c778b85f2d75d196fec

I'll create some other PRs to remove the reference and the script, thanks for the reminder.

api/v1alpha1/tiertemplaterevision_types.go
```diff
@@ -0,0 +1,33 @@
...
+// TierTemplateRevision is the Schema for the tiertemplaterevisions API
+// +kubebuilder:resource:path=tiertemplaterevisions,scope=Namespaced
+// +kubebuilder:printcolumn:name="Type",type="string",JSONPath=`.spec.type`
+// +kubebuilder:validation:XPreserveUnknownFields
+// +operator-sdk:gen-csv:customresourcedefinitions.displayName="Template Tier Revision"
+type TierTemplateRevision struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec TierTemplateSpec `json:"spec,omitempty"`
```
MatousJobanek, line 19: Do we want to use the whole TierTemplateSpec, or rather only the `spec.template`? Tier and type values will be probably stored as labels, right?

MatousJobanek (approved):
Looks OKish, but let's reconsider if we need to replicate the tier and the type, or if this is gonna be stored as labels.

api/v1alpha1/tiertemplaterevision_types.go
mfrancisc: Good point, I was thinking about this as well. From the design doc we thought about having only `spec.type` and `spec.templateObjects`, see: https://docs.google.com/document/d/1x5SoBT80df9fmVsaDgAE6DE7hE6lzmNIK087JUmgaJs/edit#heading=h.7g6qff1efnkr 

But since `spec.template` will go away later. But since `spec.TierName` and `spec.Revision` are not optional I think we should create a new spec with just:
- `spec.templateObjects`
- `spec.type` 

WDYT?

MatousJobanek: I think that in the case of the CRD it should be OK to rely on labels (for the type and the tier name) only.
Let's create spec only with the `templateObjects` for now - we can add the fields later if needed.
WDYT?

mfrancisc: makes sense - let's start simple. 

Done in https://github.com/codeready-toolchain/api/pull/446/commits/041ab3672bb51ba566a0a520a12f14a50ba14350#diff-ab1e0ce1cf613614da6482c6c0485745610c336a48d225d0f76c4f854acf5d94

MatousJobanek (approved):
Looks great, thanks

alexeykazakov (approved):

PR #450, opened 2024-11-14
Introduce ConsumedCapacity into the SpaceProvisionerConfig status

metlos:
## Description
Introduce ConsumedCapacity into the SpaceProvisionerConfig status that is meant to mirror the relevant information from the ToolchainStatus for easier inspection of the cluster conditions.

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/1104


api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -50,10 +53,28 @@ type SpaceProvisionerCapacityThresholds struct {
 	MaxMemoryUtilizationPercent uint `json:"maxMemoryUtilizationPercent,omitempty"`
 }
 
+// ConsumedCapacity describes the capacity of the cluster consumed by the spaces
+// currently provisioned to it.
+type ConsumedCapacity struct {
```
mfrancisc, line 58: minor - should we name it as in memberstatus `ResourceUsage` and keep the terminology consistent ? Since also the field `MemoryUsagePercentPerNodeRole` talks about _usage_ .

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -50,10 +53,28 @@ type SpaceProvisionerCapacityThresholds struct {
...
+	// MemoryUsagePercentPerNodeRole is the percent of the memory used per node role (eg. worker, master)
+	MemoryUsagePercentPerNodeRole map[string]int `json:"memoryUsagePercentPerNode"`
```
mfrancisc, line 60: I might be wrong but it should require also these markers
```suggestion
	// MemoryUsagePercentPerNodeRole is the percent of the memory used per node role (eg. worker, master)
	// +optional
        // +mapType=atomic
	MemoryUsagePercentPerNodeRole map[string]int `json:"memoryUsagePercentPerNode,omitempty"`
```

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -50,10 +53,28 @@ type SpaceProvisionerCapacityThresholds struct {
...
+
+	// SpaceCount is the number of spaces currently deployed to the cluster
+	SpaceCount int `json:"spaceCount"`
```
mfrancisc, line 63: same here - I usually saw that if the parent struct is optional also the fields inside are optional
```suggestion
	// SpaceCount is the number of spaces currently deployed to the cluster
	// +optional
	SpaceCount int `json:"spaceCount,omitempty"`
```

api/v1alpha1/spaceprovisionerconfig_types.go
metlos: There's `capacityThresholds` in the spec of the SPC, so `consumedCapacity` seemed likely a more logical choice to me here.

api/v1alpha1/spaceprovisionerconfig_types.go
metlos: The reason for this is that we either have a complete capacity report or none at all. Optional and omitempty only influence the JSON that is marshalled or presented to the user and since this is a status and users are not supposed to edit it, I wouldn't find it useful to provide partial capacity reports or as a user to not see "spaceCount: 0" in the status.

api/v1alpha1/spaceprovisionerconfig_types.go
metlos: I will add the `+maptype=atomic`, thanks for pointing it out. As for optional and omitempty, I don't these should be added in this case, see my reply to your comment on the `SpaceCount`. That said, I also noticed the discrepancy between the field name in go and json, so I'll fix that, too :)

api/v1alpha1/spaceprovisionerconfig_types.go
metlos: nope, we will assign the consumed capacity as a whole, not the individual fields.

mfrancisc (approved):
Looks good 👍 

Thanks for addressing my comments/questions 🙏

api/v1alpha1/spaceprovisionerconfig_types.go
```diff
@@ -50,10 +53,29 @@ type SpaceProvisionerCapacityThresholds struct {
...
+}
+
 // +k8s:openapi-gen=true
 type SpaceProvisionerConfigStatus struct {
+	// ConsumedCapacity reflects the runtime state of the cluster and the capacity it currently consumes.
+	// Nil if the consumed capacity is not known
+	// +optional
+	ConsumedCapacity *ConsumedCapacity `json:"consumedCapacity,omitempty"`
```
MatousJobanek, line 72: does it have to be a pointer?

MatousJobanek (approved):

api/v1alpha1/spaceprovisionerconfig_types.go
metlos: This is a pointer so that we can distinguish between a zero space count and unknown space count.

MatousJobanek: make sense :+1:

PR #458, opened 2025-01-20
KUBESAW-174: Print error output for repos where verification failed

fbm3307:
## Description
This PR is to update the verify dependency script such a way that it prints the error/failure in an understandable way

## Checks
1. Did you run `make generate` target? **NA**

2. Did `make generate` change anything in other projects (host-operator, member-operator)? **NA**

    - Toolchain-common: https://github.com/codeready-toolchain/toolchain-common/pull/450

### DIFF
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -9,7 +9,9 @@ ERROR_REPO_LIST=()
 ERROR_FILE_LIST=()
 STD_OUT_FILE_LIST=()
 GO_LINT_REGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
-ERROR_REGEX="Error[:]*/|FAIL[:]*" #unit test or any other failure we log from our controllers other places goes into stdoutput, hence making that regex too, to fetch the error more precisely
+ERROR_REGEX="[E\|e][R\|r][R\|r][O\|o][R\|r][:]*\|[F\|f][A\|a][I\|i][l\|L][:]*\|expected[:]*\|actual[:]*" #unit test or any other failure we log from our controllers or other places goes into stdoutput 
+                                                            #(since we log it and its not a failure in running the command or dependency check), 
+                                                            #hence making that regex too, to fetch the error more precisely
 
 echo Initiating verify-replace on dependent repos
 for repo in "${REPOS[@]}"
--- a/scripts/verify-replace.sh
+++ b/scripts/verify-replace.sh
@@ -5,10 +5,12 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e" "${GH_BASE_URL_CRT}toolchain-common")
 C_PATH=${PWD}
-ERRORREPOLIST=()
-ERRORFILELIST=()
-STDOUTFILELIST=()
-GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
+ERROR_REPO_LIST=()
+ERROR_FILE_LIST=()
+STD_OUT_FILE_LIST=()
+GO_LINT_REGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
+ERROR_REGEX="Error[:]*" #unit test or any other failure we log goes into stdoutput, hence making that regex too to fetch the error
+FAIL_REGEX="FAIL[:]*" #unit test or any other failure we log goes into stdoutput, hence making that regex too to fetch the error
 
 echo Initiating verify-replace on dependent repos
 for repo in "${REPOS[@]}"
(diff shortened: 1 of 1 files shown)


scripts/verify-replace.sh
```diff
@@ -5,7 +5,10 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e" "${GH_BASE_URL_CRT}toolchain-common")
 C_PATH=${PWD}
-ERRORLIST=()
+ERRORREPOLIST=()
+ERRORFILELIST=()
+GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\s\w]*[:]*[\s\w\">.-]*[\w\s)(*.]*"
```
mfrancisc, line 10: not an expert but should this be something like this
```suggestion
GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
```

mfrancisc (commented):
Looks good!

I have one minor comment about the regex.

scripts/verify-replace.sh
fbm3307: I had tried and tested the regex you suggested, but  we do want the specific format in which the golint displays the error. and i tried to be very specific to the format of golint and hence the above regex i formed. 
for more context, 
`Error: testsupport/wait/member.go:1321:16: G115: integer overflow conversion int -> int32 (gosec)`

`Error: testsupport/wait/awaitility.go:620:9: printf: non-constant format string in call to (*testing.common).Logf (govet)`

`Error: setup/terminal/terminal.go:49:6: the methods of "DefaultTerminal" use pointer receiver and non-pointer receiver. (recvcheck)`

the suggested regex would ideally work , but i thought of being specific to golint format so that we are sure we are not filtering anything else. WDYT ?

fbm3307: Okay i have changed the Regex to the one suggested, as i think it would be better to have a smaller regex which works. Thanks

mfrancisc: Ok, TBH I tested it with in mind this type of syntax only :
```
Error: setup/terminal/terminal.go:49:6: the methods of "DefaultTerminal" use pointer receiver and non-pointer receiver. (recvcheck)
```
I wasn't aware of the other ones. now makes more sense. I think you might be right, my version would fail for the first two.

fbm3307: The one which you suggested wont fail, it would just match the line till column number and colon, and get us the whole line during grep ..  i tested it,  it works fine .. so i just switched to it

scripts/verify-replace.sh
```diff
@@ -14,28 +18,59 @@ do
...
+                do
+                    if [[ ${s} =~ ${e} ]]; then 
+                        cat "${s}" | grep -E ${GOLINTREGEX} > lintererror
+                        if ! [ -s "lintererror" ]; then
+                            cat  lintererror   
+                        fi
+                    fi
+                done                                            
+            fi
+        done
```
mfrancisc, line 73: unless I'm missing something, the `ERRORFILELIST` is only used for matching the name of the repo and printing the header: "${e} has the following errors "

Couldn't you use the  `STDOUTFILELIST` for that instead and have only 2 for loops? Basically removing the for at line 51.

scripts/verify-replace.sh
```diff
@@ -14,28 +18,59 @@ do
...
+                echo =========================================================================================
+                echo 
+                cat "${c}"
+            else
+                for s in ${STDOUTFILELIST[*]}
+                do
+                    if [[ ${s} =~ ${e} ]]; then 
+                        cat "${s}" | grep -E ${GOLINTREGEX} > lintererror
+                        if ! [ -s "lintererror" ]; then
+                            cat  lintererror
```
mfrancisc, line 68: not sure I understand why do we need the `lintererror` variable ? 

Also , unless I'm missing something the `-s "lintererror"` is checking for file existence and I don't see where this `lintererror` file is being created . Or maybe it was supposed to be `LINTERERRORFILE` ?

scripts/verify-replace.sh
```diff
@@ -5,7 +5,11 @@ GH_BASE_URL_KS=https://github.com/kubesaw/
 GH_BASE_URL_CRT=https://github.com/codeready-toolchain/
 declare -a REPOS=("${GH_BASE_URL_KS}ksctl" "${GH_BASE_URL_CRT}host-operator" "${GH_BASE_URL_CRT}member-operator" "${GH_BASE_URL_CRT}registration-service" "${GH_BASE_URL_CRT}toolchain-e2e" "${GH_BASE_URL_CRT}toolchain-common")
 C_PATH=${PWD}
-ERRORLIST=()
+ERRORREPOLIST=()
+ERRORFILELIST=()
+STDOUTFILELIST=()
+GOLINTREGEX="[\s\w.\/]*:[0-9]*:[0-9]*:[\w\s)(*.\`]*"
+LINTERERRORFILE=$(mktemp ${TMP_DIR}LinterError.XXX)
```
mfrancisc, line 12: unless I've missed it, this doesn't seem to be used.

scripts/verify-replace.sh
fbm3307: Okay so `lintererror` file is created to just have the parsed linter error stored in it, there can be cases where the file will not be created if it doesn't have any linter error, and hence the check, then i am doing `cat lintererror` to print the errors. 
this is something as per this comment https://github.com/codeready-toolchain/toolchain-common/pull/450#discussion_r1925510222

scripts/verify-replace.sh
fbm3307: argh  i missed to use it in place of `lintererror`

scripts/verify-replace.sh
fbm3307: but just kept it for better reading of the error

fbm3307: Also `ErrorFileLIST` i am doing a cat at the end to display the error at line 61 again that is in line with this comment
https://github.com/codeready-toolchain/toolchain-common/pull/450#discussion_r1925510222

scripts/verify-replace.sh
```diff
@@ -14,28 +18,59 @@ do
...
+                echo                       "${e} has the following errors "
+                echo                                                          
+                echo =========================================================================================
+                echo 
+                cat "${c}"
+            else
+                for s in ${STDOUTFILELIST[*]}
+                do
+                    if [[ ${s} =~ ${e} ]]; then 
+                        cat "${s}" | grep -E ${GOLINTREGEX} > LINTERERRORFILE
```
metlos, line 66: I think you meant `> ${LINTERERRORFILE}` here.

scripts/verify-replace.sh
```diff
@@ -14,28 +18,59 @@ do
...
+                        if ! [ -s "LINTERERRORFILE" ]; then
```
metlos, line 67: Again you're using a literal `LINTERERRORFILE` here instead of the file in temporary directory that you store in the `$LINTERERRORFILE` in the code above.

scripts/verify-replace.sh
```diff
@@ -14,28 +18,59 @@ do
...
+                            cat  LINTERERRORFILE
```
metlos, line 68: ... and one final use of a literal instead of the variable with the same name :)

metlos (commented):
Looks good apart from some minor comments.

scripts/verify-replace.sh
fbm3307: done

scripts/verify-replace.sh
fbm3307: yes done

scripts/verify-replace.sh
fbm3307: aah yes you are right

scripts/verify-replace.sh
```diff
@@ -14,28 +18,59 @@ do
...
+                echo 
+                cat "${c}"
+            else
+                for s in ${STDOUTFILELIST[*]}
+                do
+                    if [[ ${s} =~ ${e} ]]; then 
+                        cat "${s}" | grep -E ${GOLINTREGEX} > ${LINTERERRORFILE}
+                        if ! [ -s "${LINTERERRORFILE}" ]; then
+                            cat  "${LINTERERRORFILE}"  
+                        fi
```
metlos, line 69: Looking at this code again, couldn't all this just be:
```suggestion
                        cat "${s}" | grep --color=never -E ${GOLINTREGEX}
```

fbm3307: yes, i have removed it now, 
the only reason i had added it in first place was to have temporary files containing errors, which i could use it for the  commenting purpose, but now that i think its better to remove it and see later if is required then add it.

scripts/verify-replace.sh
fbm3307: The errorreponame is actually like a central name(list with which both the stdout, errfile list can be compared) and just print the one repo at a time errors

scripts/verify-replace.sh
```diff
@@ -14,28 +20,60 @@ do
...
+        for error_file_name in ${ERROR_FILE_LIST[*]}
+        do
+            if [[ ${error_file_name} =~ ${error_repo_name} ]]; then 
+                echo                                                          
+                echo =========================================================================================
+                echo 
+                echo                       "${error_repo_name} has the following errors "
+                echo                                                          
+                echo =========================================================================================
+                echo
```
MatousJobanek, line 67:
```suggestion
        echo                                                          
        echo =========================================================================================
        echo 
        echo                       "${error_repo_name} has the following errors "
        echo                                                          
        echo =========================================================================================
        echo 
        for error_file_name in ${ERROR_FILE_LIST[*]}
        do
            if [[ ${error_file_name} =~ ${error_repo_name} ]]; then 
```

MatousJobanek (approved):

scripts/verify-replace.sh
```diff
@@ -14,28 +20,60 @@ do
...
+                echo                                                          
+                echo =========================================================================================
+                echo 
+                cat "${error_file_name}"
+            fi
+        done
+        for std_out_file_name in ${STD_OUT_FILE_LIST[*]}
+            do
+                if [[ ${std_out_file_name} =~ ${error_repo_name} ]]; then 
+                    cat "${std_out_file_name}" | grep "${GO_LINT_REGEX}\|${ERROR_REGEX}"
```
MatousJobanek, line 74:
```suggestion
                    cat "${std_out_file_name}" | grep -C 5 "${GO_LINT_REGEX}\|${ERROR_REGEX}"
```

PR #501, opened 2026-03-17
[SANDBOX-1700] chore: upgrade to OCP 4.21 dependencies

metlos:
Update openshift/api to release-4.21, k8s.io deps to v0.34.1, controller-runtime to v0.22.0, and controller-tools to v0.19.0.

https://redhat.atlassian.net/browse/SANDBOX-1700

## Checks
1. Did you run `make generate` target? **no**
This would only change `controller-gen.kubebuilder.io/version` annotation to from `0.18.0` to `0.19.0`. No actual changes to the CRDs.

3. Did `make generate` change anything in other projects (host-operator, member-operator)? **no**
See above

## Summary by CodeRabbit

* **Chores**
  * Updated core Kubernetes/OpenShift-related dependencies and supporting libraries to newer releases; refreshed several indirect dependencies and added newer YAML libraries.
  * Replaced and consolidated some Kubernetes/SIGs modules with their updated versions.
  * No public APIs or user-visible behavior changed; maintenance-focused compatibility and security updates.


alexeykazakov (approved):

go.mod
```diff
@@ -7,54 +7,56 @@ toolchain go1.24.13
 require (
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	// using latest commit from 'github.com/openshift/api branch release-4.20'
```
MatousJobanek:
```suggestion
	// using latest commit from 'github.com/openshift/api branch release-4.21'
```

MatousJobanek (approved):

go.mod
```diff
@@ -7,54 +7,56 @@ toolchain go1.24.13
...
-	github.com/openshift/api v0.0.0-20251202204302-1cb53e34ca33
-	k8s.io/api v0.33.4
-	k8s.io/apimachinery v0.33.4
-	k8s.io/code-generator v0.33.4
+	github.com/openshift/api v0.0.0-20260304172252-b0658d22beea
+	k8s.io/api v0.34.1
+	k8s.io/apimachinery v0.34.1
+	k8s.io/code-generator v0.34.1
```
MatousJobanek, line 13: out of curiosity, why not using v0.34.4 version?
Also these seems to be a mismatch between this repo and the other ones. I know that it's a z-stream, so it should be just a detail, but let's try to keep it consistent

rsoaresd: +1 regarding keeping the same version in each repo

However, I have a concern. Openshift 4.21 uses [k8s 1.34](https://github.com/openshift/api/blob/release-4.21/go.mod#L8) but [operator-sdk latest tag 1.42.1](https://github.com/operator-framework/operator-sdk/blob/v1.42.1/go.mod#L35) still uses 1.33.

According to Kanika's guideline:
> In the case that the Kubernetes version is not the same in the above two projects, then either use the smaller version of the two in which case the dependencies may not be the most up to date or wait for a version of the project that both support the same Kubernetes version.

metlos: yep, I noticed this, too. Should be fixed now and everything should be using the exact versions as the versions in OCP 4.21. That uses k8s* v0.34.1, so I wanted to be exact to limit the possibility of surprise.

go.mod
```diff
@@ -7,54 +7,56 @@ toolchain go1.24.13
...
 	k8s.io/gengo v0.0.0-20230829151522-9cce18d56c01
-	k8s.io/kube-openapi v0.0.0-20250318190949-c8a335a9a2ff
-	sigs.k8s.io/controller-runtime v0.21.0
-	sigs.k8s.io/controller-tools v0.18.0
+	k8s.io/kube-openapi v0.0.0-20250710124328-f3f2b991d03b
+	sigs.k8s.io/controller-runtime v0.22.1
```
rsoaresd: How did you "pick" this version? (Asking for learning purposes, I usually align with operator-sdk)

go.mod
```diff
@@ -6,55 +6,78 @@ toolchain go1.24.13
...
+	github.com/openshift/api v0.0.0-20260304172252-b0658d22beea
+	k8s.io/api v0.34.1
+	k8s.io/apimachinery v0.34.1
+	k8s.io/code-generator v0.34.1
 	k8s.io/gengo v0.0.0-20230829151522-9cce18d56c01
-	k8s.io/kube-openapi v0.0.0-20250318190949-c8a335a9a2ff
+	k8s.io/kube-openapi v0.0.0-20250710124328-f3f2b991d03b
 	sigs.k8s.io/controller-runtime v0.21.0
-	sigs.k8s.io/controller-tools v0.18.0
+	sigs.k8s.io/controller-tools v0.19.0
```
rsoaresd: @metlos sorry for questing a lot the versions, just asking to understand and for learning purposes as well. How did you get the `sigs.k8s.io/controller-tools` version?

metlos: I was blind, thanks for noticing this! It's been since fixed. The issue was that this was a remnant of the `controller-runtime v0.22.1` dependency tree that got left behind. I since then re-did everything straight from master to avoid these "ghosts".

go.mod
```diff
@@ -50,11 +52,11 @@ require (
...
+	k8s.io/gengo/v2 v2.0.0-20250604051438-85fd79dbfd9f // indirect
 	k8s.io/klog/v2 v2.130.1 // indirect
-	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738 // indirect
-	sigs.k8s.io/json v0.0.0-20241010143419-9aa6b5e7a4b3 // indirect
+	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397 // indirect
+	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
-	sigs.k8s.io/yaml v1.4.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
```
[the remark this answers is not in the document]
metlos: I believe this is not right. Go's module system doesn't 'resolve v4 as v6'. v4 is not present in go.mod because of dependency graph pruning.

metlos:
Postponed until operator-sdk is upgradable, too.

PR #509, opened 2026-06-04
Add PhoneLookupMode config and phone-lookup annotations

fbm3307:
- Add PhoneLookupMode (disabled/log/enabled) to RegistrationServiceVerificationConfig

related Prs:
Registration-Service: https://github.com/codeready-toolchain/registration-service/pull/595
toolchain-Common: https://github.com/codeready-toolchain/toolchain-common/pull/532
host operator: https://github.com/codeready-toolchain/host-operator/pull/1268
toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/1285

## Description
- Add PhoneLookupMode (disabled/log/enabled) to RegistrationServiceVerificationConfig
- Add six phone-lookup annotation key constants to UserSignup
- Regenerate deepcopy, openapi, CRD docs

Part of the Twilio Lookup api feature — see [sandbox-sre#3192]

## Checks

    - host-operator: https://github.com/codeready-toolchain/host-operator/pull/1268
  
  Assisted by : Cursor

## Summary by CodeRabbit

* **New Features**
  * Added a configurable phoneLookupMode for phone verification (disabled, log, enabled; default: log).
  * Option to exclude specific countries from phone lookup checks.
  * Capture phone lookup metadata on signups: carrier risk category, number-blocked flag, and combined lookup details.

* **Documentation**
  * API reference updated to describe the new setting, allowed values, default, and exclusion behavior.


metlos (approved):

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -386,6 +386,14 @@ type RegistrationServiceVerificationConfig struct {
...
 	// +listType=atomic
 	TwilioSenderConfigs []TwilioSenderConfig `json:"twilioSenderConfigs,omitempty"`
+
+	// PhoneLookupMode controls how the registration service handles Twilio Lookup v2 phone risk checks.
+	// Valid values are "disabled" (skip Lookup entirely), "log" (call Lookup and store results but don't block),
+	// and "enabled" (call Lookup and enforce blocking). Defaults to "log".
+	// +optional
+	// +kubebuilder:default="log"
+	// +kubebuilder:validation:Enum=disabled;log;enabled
+	PhoneLookupMode *string `json:"phoneLookupMode,omitempty"`
```
metlos: What about making this a true string-based enum? I didn't realize this at first but seeing a stringly-typed argument in toolchain-common and elsewhere makes me want it :)

fbm3307: updated it

api/v1alpha1/usersignup_types.go
```diff
@@ -52,6 +52,21 @@ const (
...
+	UserSignupPhoneLookupNumberBlockedAnnotationKey = LabelKeyPrefix + "phone-lookup-number-blocked"
+	// UserSignupPhoneLookupRiskScoreAnnotationKey stores the sms_pumping_risk_score
+	UserSignupPhoneLookupRiskScoreAnnotationKey = LabelKeyPrefix + "phone-lookup-risk-score"
+	// UserSignupPhoneLookupCarrierNameAnnotationKey stores the carrier name from line_type_intelligence
+	UserSignupPhoneLookupCarrierNameAnnotationKey = LabelKeyPrefix + "phone-lookup-carrier-name"
+	// UserSignupPhoneLookupLineTypeAnnotationKey stores the line type (mobile, landline, voip, etc.)
+	UserSignupPhoneLookupLineTypeAnnotationKey = LabelKeyPrefix + "phone-lookup-line-type"
+	// UserSignupPhoneLookupPhoneHashAnnotationKey stores the MD5 hash of the phone number that was last looked up (for retry optimization)
+	UserSignupPhoneLookupPhoneHashAnnotationKey = LabelKeyPrefix + "phone-lookup-phone-hash"
+
```
MatousJobanek: such a list of annotations seems to be an overkill to me:
* let's pick one or two which are the most important ones and store them (looking at reg-service PR, it's the `phone-lookup-carrier-risk` and `phone-lookup-number-blocked`
* the rest can be either logged or if they are  important to keep them, then store them as a single-line Json value in one annotation

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -386,6 +396,13 @@ type RegistrationServiceVerificationConfig struct {
 	// +optional
 	// +listType=atomic
 	TwilioSenderConfigs []TwilioSenderConfig `json:"twilioSenderConfigs,omitempty"`
+
+	// PhoneLookupMode controls how the registration service handles Twilio Lookup v2 phone risk checks.
+	// Valid values are "disabled" (skip Lookup entirely), "log" (call Lookup and store results but don't block),
+	// and "enabled" (call Lookup and enforce blocking). Defaults to "log".
+	// +optional
+	// +kubebuilder:default="log"
+	PhoneLookupMode *PhoneLookupMode `json:"phoneLookupMode,omitempty"`
```
MatousJobanek, line 405: Looking at reg-service, we will need also a list of countries to be excluded (Canada for now)

MatousJobanek: see https://github.com/codeready-toolchain/sandbox-sre/pull/3192#discussion_r3371345352

api/v1alpha1/usersignup_types.go
fbm3307: updated it

api/v1alpha1/toolchainconfig_types.go
fbm3307: updated it

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -386,6 +396,20 @@ type RegistrationServiceVerificationConfig struct {
...
+
+	// PhoneLookupExcludedCountries is a list of ISO 3166-1 alpha-2 country codes (e.g. ["CA", "US"])
```
MatousJobanek, line 407: what should be set here? the country codes `CA` `US` or rather the country code numbers eg: `1` - in reg-service we have the country code numbers, not the abbreviation. we also use numbers in 
https://github.com/codeready-toolchain/api/blob/cfa3dd9db9cc8464d25b37a96017529b09cad369/api/v1alpha1/toolchainconfig_types.go#L404-L407

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -386,6 +396,20 @@ type RegistrationServiceVerificationConfig struct {
...
+	// for which Twilio Lookup should be skipped. Twilio Lookup data quality is insufficient
+	// for these regions. Canada also has CLNPC regulatory restrictions.
```
MatousJobanek:
```suggestion
	// for which Twilio Lookup should be skipped.
```

api/v1alpha1/toolchainconfig_types.go
```diff
@@ -386,6 +396,20 @@ type RegistrationServiceVerificationConfig struct {
...
+	// +optional
+	// +listType=atomic
```
MatousJobanek: `set` is probably better type in this case
```suggestion
	// +listType=set
```

api/v1alpha1/usersignup_types.go
```diff
@@ -52,6 +52,14 @@ const (
 	// The time is stored in time.RFC3339 format, the reason is that the main purpose of the metric is to track the UX of the users, and milliseconds don't really matter in that case
 	UserSignupRequestReceivedTimeAnnotationKey = LabelKeyPrefix + "request-received-time"
 
+	// UserSignupPhoneLookupCarrierRiskAnnotationKey stores the carrier_risk_category from Twilio Lookup (e.g. "low", "moderate", "high")
+	UserSignupPhoneLookupCarrierRiskAnnotationKey = LabelKeyPrefix + "phone-lookup-carrier-risk"
+	// UserSignupPhoneLookupNumberBlockedAnnotationKey stores whether the number is blocked ("true"/"false")
+	UserSignupPhoneLookupNumberBlockedAnnotationKey = LabelKeyPrefix + "phone-lookup-number-blocked"
+	// UserSignupPhoneLookupDetailsAnnotationKey stores a single-line JSON value with supplementary lookup data
+	// (result, risk_score, carrier_name, line_type, phone_hash)
```
MatousJobanek: unneccessarry implementation detail
```suggestion
```

api/v1alpha1/toolchainconfig_types.go
fbm3307: we need to use "US"/"CA". We need ISO alpha-2 codes here, not calling code numbers. The reason is that country calling code "1" is the NANP (North American Numbering Plan) code shared by 20+ countries — US, Canada, Jamaica, Puerto Rico, Dominican Republic, Trinidad, Barbados, etc.

If we use "1" to exclude countries, we'd skip Lookup for all NANP countries, not just US and Canada. With ISO codes ("US", "CA"), we can exclude them individually while still running Lookup for +1-876-xxx (Jamaica) or +1-809-xxx (Dominican Republic).

The existing CountryCodes field in TwilioSenderConfigs uses calling code numbers because it's matching on the dialing prefix the user submits (for choosing which Twilio sender/route to use). That's a different use case — there, matching on "1" to pick a sender for all NANP numbers is correct.

Here, we're controlling which countries to skip Lookup for, so we need per-country granularity. We use Google's phonenumbers library to resolve the full E.164 number to the ISO region: +1-415-xxx → "US", +1-613-xxx → "CA", +1-876-xxx → "JM".

MatousJobanek: ok, but why did you write in reg-service check by number? https://github.com/codeready-toolchain/sandbox-sre/pull/3192#discussion_r3371345352

MatousJobanek: it looks like that you changed that in reg-service, I cannot find that anymore. TBH, trying to orientate in the bunch of changes across all the PRs is a bit challenge. 

This type of information should be in the spec clearly visible, when you introduce a change like this, you should mention that in the comment so the reviewer doesn't miss it.
Also, I would prefer having spec WITHOUT implementation details, but WITH important information like this one.

fbm3307: apologize for the confusion, it was a miss from my part, there was a initiall implementation with number, when you commented, thats when i found out that it cannot be done with numbers and hence i updated the logic and spec all together, 

i will keep that in mind for the next time to sort the spec first and then start with implementation with clarity

MatousJobanek (approved):

